PR #15, opened 2017-12-12
Add initial number-insight OAS3 definition

adambutler:
Adds Open API Specification 3 definition for Number Insight.

- [x] Document endpoints
- [x] Investigate use of `niResponse` & `oneOf` and possible other renderer friendly options
- [x] Add example values for all parameters
- [x] Use `enum` to provide possible values. Remove possible values from descriptions.
- [x] Investigate implementing a vender extension to provide descriptions for possible values.

### DIFF
--- a/number-insight.yml
+++ b/number-insight.yml
@@ -183,6 +183,37 @@ components:
     format:
       type: object
       properties:
+        lookup_outcome:
+          type: object
+          properties:
+            code:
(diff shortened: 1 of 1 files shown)


adambutler:
I'll also merge this as-is. If further tweaks are required please open a PR against `master`.

PR #16, opened 2017-12-13
Verify API - WIP

MikeRalphson:
As discussed, may need the following resolved:

* ~~I assume the request:next_event_wait parameter is in seconds?~~ Yes
* ~~The API appears to always return status code 200, even for incorrect credentials. Are there any other response codes to take into account? E.g. do you want to advertise an explicit '5xx' response code for server issues?~~ Agreed 200 is sufficient.
* ~~The API appears to return a Content-Type of 'text/plain' when format:xml is requested. Should the OpenAPI response.content keys match the delivered Content-Type (text/plain) or the actual type of the response (text/xml or similar)?~~ Agreed: document API as-is with `text/plain`
* ~~Do you want a pattern property against IP addresses, and are both IPv4 and IPv6 addresses supported? Happy to add for completeness, but I think the pattern property would be quite ugly.~~ Will be looked at in the future.
* Couldn't spot any response headers (e.g. rate limiting) which looked like they needed documenting, but please let me know on this.

### DIFF
--- a/verify.yml
+++ b/verify.yml
@@ -318,6 +318,7 @@ paths:
             items:
               type: string
               example: aaaaaaaafffffffff0000000099999999
+            maxItems: 10
           style: form
           explode: true
       responses:
--- a/verify.yml
+++ b/verify.yml
@@ -108,7 +108,7 @@ paths:
           schema:
             type: string
             maxLength: 18
-          example: MyApp
+          example: Acme Inc
         - name: sender_id
           required: false
           in: query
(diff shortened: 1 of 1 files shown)


adambutler:
@MikeRalphson I'm happy that this provides a great starting point so I am going to merge since I know of no known issues.

Given our conversation, I'm under the assumption that your not actively working on this spec at the moment but if you do have any further tweaks to make please do so by opening a new PR against master.

I'll look over the coming days about implementing these into [Nexmo Developer](https://developer.nexmo.com) and will report back if I find any issues on the way and make a list of recommendations for defining the remaining APIs if appropriate.

adambutler:
I had to move this over to another PR due to what seems to be a GitHub bug handling permissions.

See #17 for more information.

PR #19, opened 2017-12-19
Add cursor pagination example

mheap:
After a chat with @afolson we came to the conclusion that cursor based navigation is the best course of action going forwards. This is for three main reasons:

* VAPI already (kind of) uses cursor based navigation (via `record_index`)
* A cursor means no duplicated responses and no missing responses for the user
* Using json/hal and `_links.next` the pagination logic is taken away from the consumer. They just follow the link provided

### Missing data example

Imagine that we have following data set:

```json
[
  {
    "id": 1,
    "text": "one"
  },
  {
    "id": 2,
    "text": "two"
  },
  {
    "id": 3,
    "text": "three"
  },
  {
    "id": 4,
    "text": "four"
  }
]
```

We request this data with `page_size == 2` and `page_index == 0`, returning the following data:

```json
[
  {
    "id": 1,
    "text": "one"
  },
  {
    "id": 2,
    "text": "two"
  }
]
```

Then, between that and our next request (`page_size == 2` and `page_index == 1`) ID 2 is deleted. Our request will return the following:

```json
[
  {
    "id": 4,
    "text": "four"
  }
]
```

ID 3 is never seen as now it belongs to the first page, but we've already accessed that page so we miss it.

The same example with a cursor (`page_size == 2` and `cursor == 2`)

```
[
  {
    "id": 3,
    "text": "three"
  },
  {
    "id": 4,
    "text": "four"
  }
]
```

This works as we tell the API that the latest ID we saw was 2, so show anything newer than that (even if 2 doesn't exist any more).

### Cons of using a cursor

In general, pagination via `page_size` and `page_index` is preferable for users not familiar with APIs. Cursors are designed mainly for high throughput websites (e.g. Twitter and Slack both use cursors).

However, as we provide the next page via `_links.next`, the user just has to follow the link we provide to get the next page of results. This removes the need for them to understand how cursors work meaning that the main disadvantage of cursors is removed.

### DIFF
--- a/guidelines/GUIDELINES.md
+++ b/guidelines/GUIDELINES.md
@@ -278,16 +278,104 @@ All new API's shall adhere to the principles of REST (http://www.ics.uci.edu/~fi
 - `DELETE members/steve` - remove the member Steve from the collection
 
 ### Pagination
+
+APIs at Nexmo shall use one of the following pagination schemes:
+
+* Cursor based for ever-expanding data sets, or where it's infeasible to count
+the total number of records e.g. Voice API call log. Cursor based
+pagination is also used when data is programatically added to a data set e.g.
+call recordings to the Media API
+* Page based for small, managed APIs where the customer is in control of what
+is in the data set e.g. Number pools
+
+To the customer, having two different pagination methods is OK, as in practice
+they should request the first page and then follow the `_links.next` URL to get
+the next page.
+
+Cursor implementation can be decided by the engineering team. Results returned
+must be consistently ordered, and the customer can decide if the data set
+is returned in ascending or descending order.
+
+#### Cursor based
+
+Nexmo APIs shall use a cursor as their pagination option unless otherwise
+agreed with DevRel.
+
+> Cursors, if you’re not familiar, are like pointers. Pointers point at things,
+they reference a specific iota, a place in the list where your last request left
+off. A bookmark with breadcrumbs built in. (via [Slack](https://api.slack.com/docs/pagination#cursors))
+
 - There shall be a way to page through collections without additional filters.
 - There shall be a way to page through collections with filters.
+- Cursors shall not expire
 - Paging shall use a standard set of hal+json `_links`
   - `self` (current page, required)
   - `next` (next page, optional)
   - `prev` (previous page, optional)
   - `first` (first page, required)
   - `last` (last page, required)
-- Paging `_links` will include filters.
+- Paging `_links` must include filters.
+- The page size parameter must be called `page_size`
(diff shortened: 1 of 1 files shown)


guidelines/GUIDELINES.md
```diff
@@ -287,6 +287,28 @@ All new API's shall adhere to the principles of REST (http://www.ics.uci.edu/~fi
   - `first` (first page, required)
   - `last` (last page, required)
 - Paging `_links` will include filters.
+- Paging will be done via a cursor, not a page offset
+
+```json
+{
+  "page_size": 100,
+  "cursor": "19284743",
```
leggetter: Is this a replacement for `page_index` that has been used in other examples? Or `record_index` in the Voice API?

mheap: It's closer to a replacement for `record_index` but I still don't understand how `record_index` works based on this statement

> Return page_size calls from this index in the response. That is, if your request returns 300 calls, set record_index to 5 in order to return calls 50 to 59. The default value is 0. That is, the first page_size calls.

I interpreted it as:

> Set `record_index` to 5, we'll multiply that by 10 and start returning `page_size` results from 50. If you set `record_index` to 7 we'll return `page_size` results starting from 70

This is *almost* like a cursor (decoupled `page_size` and `offset`, but not quite as there's still some implicit multiplication.

I'd have to play with VAPI some more to work out it's existing behaviour.

arisgoulia:
Looks good to me.

tommorris:
The only thing I'd suggest here is to make it clear that cursors are strings and not integers. They should be interpreted as arbitrary rather than having any semantic meaning. The implementation detail is irrelevant—it could be something like a cryptographic hash (SHA-256 etc.) or similar. Or you could go to the next step and just make it a URL.

All the consumer of the API needs to know is "when you want more content, just follow its nose to the next bit".

An example in widespread use is the MediaWiki API. If you look at [this example](https://en.wikipedia.org/w/api.php?action=query&list=categorymembers&cmtitle=Category:Films_about_dragons) (prettified) JSON response, you can see a key called `continue`. `continue` contains a key called `cmcontinue` (`cm` being "category member")—the key is the URL parameter to use in the followup request and the value of it is to be passed as the value of that URL parameter. With the MediaWiki example, you know when you have exhausted content as the `continue` data no longer exists in the response.

The nice thing about this kind of design is languages like Python can abstract away the API calls (and even rate limiting) using a [generator](https://wiki.python.org/moin/Generators). As an API consumer, you get back a data structure that keeps giving you data you can iterate through but will also load more data on request.

PR #20, opened 2018-01-05
Adding Reporting API

tommorris:


definitions/reporting.yml
```diff
@@ -0,0 +1,355 @@
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/NewReportResp'
+        '400':
+          description: Bad Request
+        '401':
+          description: Unauthorized
+        '403':
+          description: Forbidden
+        '500':
```
adambutler: Not sure why we would document `500` since it is not a known condition?

As for `400`, `401`, `403` if they are producible responses we should be documenting them with the appropriate schema for that response code.

definitions/reporting.yml
```diff
@@ -0,0 +1,355 @@
...
+          description: Your API secret. Required unless `sig` is provided
+          required: false
+          in: query
+          example: abcdef0123456789
+          schema:
+            type: string
+            minLength: 16
+            maxLength: 16
+        - name: sig
+          description: The hash of the request parameters in alphabetical order, a timestamp and the signature secret. See [Signing Requests](/concepts/guides/signing-messages) for more details.
```
adambutler: The URL in the description should be absolute since schema can be used externally.

definitions/reporting.yml
```diff
@@ -0,0 +1,355 @@
...
```
adambutler: The URL in the description should be absolute since schema can be used externally.

definitions/reporting.yml
```diff
@@ -0,0 +1,355 @@
...
+            type: string
+            minLength: 16
+            maxLength: 16
+        - name: file_id
+          in: path
+          schema:
+            type: string
+            example: aaaaaaaa-bbbb-cccc-dddd-0123456789ab
+          description: UUID of the file.
+          required: true
```
adambutler: Check if `required` is valid for `path` type.

definitions/reporting.yml
```diff
@@ -0,0 +1,355 @@
...
+          description: Your API secret. Required unless `sig` is provided.
+          required: false
+          in: query
+          example: abcdef0123456789
+          schema:
+            type: string
+            minLength: 16
+            maxLength: 16
+        - name: sig
+          description: The hash of the request parameters in alphabetical order, a timestamp and the signature secret. See [Signing Requests](/concepts/guides/signing-messages) for more details.
```
adambutler: The URL in the description should be absolute since schema can be used externally.

definitions/reporting.yml
```diff
@@ -0,0 +1,355 @@
...
+        account_ref:
+          type: string
+          description: For MT SMS reports - search only for messages with this `account_ref`.
+          maxLength: 40
+          example: myaccountref
+        account_id:
+          type: string
+          description: Account ID for the report.
+          example: abcde
+        include_subaccounts:
```
adambutler: Is this really a string? Has it been raised as a design issue?

definitions/reporting.yml
tommorris: Fixed. (Will need to be fixed in `sms.yml`.)

definitions/reporting.yml
tommorris: Fixed.

definitions/reporting.yml
tommorris: Fixed.

definitions/reporting.yml
tommorris: Yes, an account ID is either the primary account identifier (same as the API key), or the subaccount identifier (which is the same as the subaccount's API key) - an eight character hex number. It absolutely should be a string.

definitions/reporting.yml
tommorris: To quote [OAS 3.0.1](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#parameterObject):

> If the parameter location is "path", this property is **REQUIRED** and its value MUST be `true`. Otherwise, the property **MAY** be included and its default value is `false`.

definitions/reporting.yml
tommorris: Done all three in a slightly clumsy way since the 400, 401 and 403 error codes have inconsistent JSON formatting. Hopefully we can merge these into one unified error message format when that has been fixed.

Removed the 500.

lornajane:
This definition was added to NDP and will be migrated from its current version there into this repo.

PR #22, opened 2018-01-09
number-insight; rework response objects

MikeRalphson:
* Separate models for niResponseXml and niResponseJson
* Removed niDetail / niError / format / response

There appears to be no way to convince swagger-editor/-ui that the xml element name of a root schema can vary. An alternate representation is:

```yaml
    niResponseXml:
      type: object
      description: 'The root object of the XML response varies based on the detail level requested.'
      allOf:
        - oneOf:
          - xml:
              name: 'format'
          - xml:
              name: 'lookup'
        - oneOf:
          - $ref: '#/components/schemas/niResponseXmlFormat'
          - $ref: '#/components/schemas/niResponseXmlLookup'
```
  
Though this does not tie the `xml:name`s to the corresponding structures tightly.


definitions/number-insight.yml
```diff
@@ -164,402 +163,352 @@ components:
...
+                  description: 'Information about the provided IP address'
+                  $ref: '#/components/schemas/niIP'
+                ip_warnings:
+                  type: string
+                  description: 'Warning levels for `ip`'
+                  enum:
+                    - unknown
+                    - no_warning
+                  example: 'no_warning'
+                caller_identity:
```
MikeRalphson: Not sure of provenance of `caller_identity` as a separate object / whether needs special treatment for Xml format. Description appears to be copied from `roaming`?

adambutler:
@MikeRalphson Thanks for sending these changes. I just wanted to let you know I am looking into this PR.

I intended to be able to review this today but unfortunately I'm still trying to get my head around it. In regards to the root name maybe a vendor extension of something `x-xml-root-name` might be our only option. What do you think?

I'll take another look tomorrow morning and I'll respond here as soon as I can.

adambutler:
Sorry for the slow reply, it was a bit more complicated than I anticipated and the team were also away for a week. I've pushed up a bunch of changes to this definition, I'll provide a bit more context here:

5222b2d Copy tweaks - Just as it sounds is just minor tweaks to the copy. Contains no structural changes.

335f946 Remove 401 response - This API does not respond with a 401 when invalid credentials are given so this has been removed.

cf37dd6 Fix description in wrong place - Moved one description key that was in the wrong place.

39ff588 Add async schema - Async endpoint was previously documented to respond with `niResponse`. This change corrects that and provides a new schema for the actual response of a call to the async path.

d2ec1fb Split out schemas, fix issues - Too many inconsistencies were found between JSON & XML that I believe that changing the schema to separate out these possible responses was for the best. The structure is now:

```
- niResponseXml
  - niResponseXmlBasic
  - niResponseXmlStandard
  - niResponseXmlAdvanced
- niResponseJson
  - niResponseJsonBasic
  - niResponseJsonStandard
  - niResponseJsonAdvanced
```

Furthermore, some inconsistencies such as `roaming` always having a static schema for standard has meant that duplication vs inheriting properties provides the simplest, accurate solution at this time. This could potentially be improved to reduce duplication in the future by moving the consistent properties into components that can be included.

6b7c76a Add x-group for Number Insight - Adds a group for Number Insight (Nexmo Developer specific).
8bc5b43 Add code example pointers for NI - Adds code example reference into document (Nexmo Developer Specific).

Lastly, just a quick aside on `xml` options:

### `xml.attribute`

Support in Nexmo Developer has been added here so that:

```json
{
  "type": "object",
  "properties": {
    "code": {
      "type": "integer",
      "xml": {
        "attribute": true
      }
    },
    "foo": {
      "type": "object",
      "properties": {
        "foo": {
          "type": "integer",
          "xml": {
            "attribute": true
          }
        }
      }
    }
  }
}
```

Represents:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<hash code="1">
  <foo foo="1">
  </foo>
</hash>
```

### `xml.name`

Support in Nexmo Developer has been added here so that:

```json
{  
  "type":"object",
  "properties":{  
    "foo_bar":{  
      "type":"string",
      "xml":{  
        "name":"fooBar"
      }
    }
  }
}
```

Represents:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<hash>
  <fooBar>abc123</fooBar>
</hash>
```

> Note: this also resolves the root node issue that we discussed previously.

### `xml.text` / `xml.x-text`

Lastly following your proposal in this [issue](https://github.com/OAI/OpenAPI-Specification/issues/630). I've added support for both `text` and `x-text` to be supported. I wanted to confirm with you that this behaves how you intended it to:

```json
{  
  "type":"object",
  "properties":{  
    "foo":{  
      "type":"object",
      "properties":{  
        "bar":{  
          "type":"string",
          "xml":{  
            "text":true
          }
        }
      }
    }
  }
}
```

Represents:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<hash>
  <foo>abc123</foo>
</hash>
```

But when JSON would represent:

```json
{
  "foo": {
    "bar": "abc123"
  }
}
```

---

You can see this definition in practice at https://nexmo-developer-staging-pr-545.herokuapp.com/api/number-insight.

Hopefully with these issues out of the way documenting the remaining APIs will be a bit faster and easier going forward. I'd propose where there are may be many different responses or formats that we:

1. Document the simplest JSON response first
2. Evaluate how different the XML response / other possible responses are in comparison
3. Decide if splitting schemas or documenting within the same schema is most appropriate.
4. Complete the definition either split or unified.

adambutler:
I've made a few more tweaks and fixed. I'm pretty happy with the definition now and will merge. Any further tweaks should be opened as another PR.

PR #26, opened 2018-02-16
caller_name appears for both individual & business

tommorris:
Fixes JIRA DOCS-260


tommorris:
Approving as it has been awaiting review for too long.

PR #30, opened 2018-03-22
Update sms.yml

sammachin:
corrected the length of a phone number in to: we require e.164 numbers which the specification states has a max length of 15 digits, IIRC the largest in current use is 13 digits.
I can't see a min length in the spec but the shortest length in use is Nieu with a 7 digit e.164 number (3 digit cc and 4 digit national number so that feels like a safe bet


adambutler:
Looks good to me, please bump the patch version in version.rb and 🚀

PR #32, opened 2018-04-04
Add developer-messages OpenAPI definition

MikeRalphson:
First cut of developer-messages API definition

* Includes only `200` responses for now as per @adambutler 
* Avoids use of `oneOf` in response schemas (https://github.com/swagger-api/swagger-ui/issues/3803#issuecomment-377905800)
* Response media-types are as returned by the API (includes `charset`)

### DIFF
--- a/definitions/developer-messages.yml
+++ b/definitions/developer-messages.yml
@@ -54,10 +54,10 @@ paths:
         '200':
           description: OK
           content:
-            application/json;charset=UTF-8:
+            application/json:
               schema:
                 $ref: '#/components/schemas/message'
-            application/xml;charset=UTF-8:
+            application/xml:
               schema:
                 $ref: '#/components/schemas/message'
   /messages:
@@ -102,10 +102,10 @@ paths:
         '200':
           description: OK
           content:
-            application/json;charset=UTF-8:
+            application/json:
               schema:
                 $ref: '#/components/schemas/messages'
-            application/xml;charset=UTF-8:
+            application/xml:
               schema:
                 $ref: '#/components/schemas/messages'
   /rejections:
@@ -139,10 +139,10 @@ paths:
         '200':
           description: OK
           content:
-            application/json;charset=UTF-8:
+            application/json:
               schema:
                 $ref: '#/components/schemas/rejections'
-            application/xml;charset=UTF-8:
+            application/xml:
               schema:
                 $ref: '#/components/schemas/rejections'
 components:
--- a/definitions/developer-messages.yml
+++ b/definitions/developer-messages.yml
@@ -95,6 +95,7 @@ paths:
         - name: ids
           in: query
           description: The list of up to 10 message IDs to search for.
+          x-nexmo-developer-collection-description-shown: true
           style: form
           explode: true
           schema:
@@ -102,6 +103,7 @@ paths:
             items:
               type: string
               description: A message ID to search for.
+              x-nexmo-developer-collection-description-shown: true
               example: 00A0B0C0
             minItems: 1
             maxItems: 10
--- a/definitions/developer-messages.yml
+++ b/definitions/developer-messages.yml
@@ -1,0 +1,477 @@
+openapi: 3.0.0
+info:
+  title: Nexmo Messages API
+  version: 1.0.0
+  description: >-
+    The Messages API lets you retrieve messages you have sent via the SMS API by
+    ID, as well as retrieve details of messages that were rejected.
+  contact:
+    name: Nexmo.com
+    email: devrel@nexmo.com
+    url: 'https://developer.nexmo.com/'
+    x-twitter: Nexmo
+  termsOfService: 'https://www.nexmo.com/terms-of-use'
+  license:
+    name: The MIT License (MIT)
+    url: 'https://opensource.org/licenses/MIT'
+  x-logo:
+    url: 'https://twitter.com/Nexmo/profile_image?size=original'
+  x-apiClientRegistration: 'https://dashboard.nexmo.com/sign-up'
+servers:
+  - url: 'https://rest.nexmo.com/search'
+externalDocs:
+  url: 'https://developer.nexmo.com/api/developer/messages'
+  x-sha1: d8836c374e2a7504bd2cd59e05fcee440f67cb44
+security:
+  - apiKey: []
+    apiSecret: []
+tags:
+  - name: Search
+    description: Search.
+  - name: Retrieve Multiple
+    description: Retrieve multiple messages.
+  - name: Retrieve Rejected
+    description: Retrieve rejected messages.
+paths:
+  /message:
+    get:
+      description: >-
+        Retrieve information about a single message that you sent using SMS API
+        or that was received on your number.
+      operationId: searchMessage
+      tags:
+        - Search
+      parameters:
+        - name: id
+          in: query
+          required: true
+          description: The ID of the message you want to retrieve.
+          example: 00A0B0C0
+          schema:
+            type: string
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json;charset=UTF-8:
+              schema:
+                $ref: '#/components/schemas/message'
+            application/xml;charset=UTF-8:
+              schema:
+                $ref: '#/components/schemas/message'
+  /messages:
+    get:
+      description: >-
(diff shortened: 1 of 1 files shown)


definitions/developer-messages.yml
```diff
@@ -0,0 +1,477 @@
...
+          required: true
+          description: The ID of the message you want to retrieve.
+          example: 00A0B0C0
+          schema:
+            type: string
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json;charset=UTF-8:
```
adambutler: Thoughts on changing this to just be `application/json`?

definitions/developer-messages.yml
```diff
@@ -0,0 +1,477 @@
...
+              schema:
+                $ref: '#/components/schemas/message'
+            application/xml;charset=UTF-8:
```
adambutler: 1. Should this not be `text/xml` instead of `application/xml`?
2. Again thoughts on removing `;charset=UTF-8`?

definitions/developer-messages.yml
```diff
@@ -0,0 +1,477 @@
...
+              schema:
+                $ref: '#/components/schemas/message'
+  /messages:
+    get:
+      description: >-
```
adambutler: Can we add `summary` here too please. In this case "Search Messages"

definitions/developer-messages.yml
```diff
@@ -0,0 +1,477 @@
...
+        - name: ids
+          in: query
+          description: The list of up to 10 message IDs to search for.
+          example: 00A0B0C0
+          style: form
+          explode: true
+          schema:
+            type: array
+            items:
+              type: string
```
adambutler: Not sure this is structured quite right. It could be a bug in my renderer but this is the output I'm getting:

nexmo_developer

definitions/developer-messages.yml
adambutler: ℹ️ Sidenote for anyone who looks at this in the future ...

Either the render or parser doesn't seem to like this value `application/json;charset=UTF-8`.

It's likely that there is some sort of condition for example:

```ruby
if content_type == 'application/json'
  # do thing
end
```

Assuming that is the case if this value should stay as it this could be changed to something like:

```ruby
if content_type.include? 'application/json'
  # do thing
end
```

MikeRalphson: The full value is what is currently being returned by the live API. Some API testing tools (such as Apiary's Dredd - though that is currently OpenAPI 2.0 only) expects the `Content-Type` header to match the media-type object key exactly, incuding the media-type parameters such as `charset`. For simplicity of documentation, I'm happy to change this to `application/json`.

Re: the parser/renderer issue, elsewhere I am using the regex `^(application|text){1}\\/(.*\\+){0,1}json(;){0,1}(\\s){0,}(charset=.*){0,}$`

definitions/developer-messages.yml
MikeRalphson: Again, looking at my notes that's what is being returned by the live API. Happy to change to `text/xml` to show intent though.

definitions/developer-messages.yml
MikeRalphson: Will do.

definitions/developer-messages.yml
MikeRalphson: Thanks - will take another look.

MikeRalphson:
> With regards to "Avoids use of oneOf in response schemas" we actually support this in our renderer 

Good to hear. I will try and get the renderer running in docker so I can test that locally.

> Regardless I'm interested in getting some details about how you 'avoided' it.

Partially because of ignoring the non-200 responses, I was able to rework my initial take, which looked like this:

```yaml
components:
  schemas:
    message:
      oneOf:
      - type: object
        properties:
          error-code:
            $ref: '#/components/schemas/errorCode'
          error-code-label:
            $ref: '#/components/schemas/errorCodeLabel'
      - $ref: '#/components/schemas/messageBase'
      - allOf:
        - $ref: '#/components/schemas/messageBase'
        - $ref: '#/components/schemas/messageExtension'
```

I'll see if it needs to be reinstated if we add the error responses.

The only 'gotcha' in there at the moment is:

```yaml
        items:
          type: array
          xml:
            wrapped: true
          items:
            allOf:
              - $ref: '#/components/schemas/message'
              - xml:
                  name: message
```

Which convinces at least swagger-editor/swagger-ui to name the xml elements correctly, and gets around the fact that sibling properties of `$ref` are ignored.

definitions/developer-messages.yml
MikeRalphson: I'll add a description for the array items (and move the example, which was a bug). FWIW swagger-ui/editor do render the description for `ids`:

image

I'm not sure where the `Key` of `type` is coming from though - as the renderer appears to know it's an array of strings.

MikeRalphson:
I now have it on good authority that `application/json` counts as a media-type **range** so this description from OAS 3 applies:

> The key is a media type or media type range and the value describes it. For responses that match multiple keys, only the most specific key is applicable. e.g. text/plain overrides text/*

So `application/json` **would** match `application/json;charset=UTF-8` responses.

Where the API actually generates `application/xml` not `text/xml` I think the OpenAPI definition should follow this, otherwise we would have to include both media-types or use `*/xml` which to my mind is confusing.

definitions/developer-messages.yml
ChrisGuzman: @MikeRalphson did we resolve this? I'm still seeing the same error that @adambutler pointed out above ^

ChrisGuzman:
Hi @MikeRalphson 

Sorry for jumping in here without that much knowledge. Is this PR still a WIP?:

Are their plans to fill in some of the missing features on the current documentation page? 
https://developer.nexmo.com/api/developer/messages/json

Comparing and contrasting these documents, the OAS doc in this PR is missing the following: 

- [ ] XML response objects for all of the endpoints.
- [ ] The response for Inbound Message (MO)
- [ ] Fields for MT messages only

Another question:
This spec notes that api key and secret are not required to make api calls to the Nexmo Messages API. Is that correct?

Are there any questions you have for me?

definitions/developer-messages.yml
MikeRalphson: Yes it was amended in 6d80a15 to move the `example` into the `items` and provide a `description` at both the `array` and `items` level. Not quite sure why that would not be rendering? Is there another example of `array` parameters in another definition which does render correctly?

MikeRalphson:
Hi @ChrisGuzman - yes this PR is still a WIP, for example pending a decision on whether `400` / `401` responses need including.

In swagger-editor / swagger-ui at least, I can see that `application/xml` mediatypes are present for each response (`xml` annotations only being necessary where the xml element name is different from the json property name) and that the combination of `api_key` / `api_secret` is defined at the API level and thus applies to each operation.

I can split the model for MO/MT at the cost of some more `$ref`s to avoid duplication, no problem.

ChrisGuzman:
Ok let me see if I can get the xml to show in Nexmo's OAS parser, it does seem to be like a bug on the parser side. I'll find out about including 400 / 401 responses, but I believe the answer is going to be yes.

ChrisGuzman:
Yes let's document that 400 and 401 responses

MikeRalphson:
@ChrisGuzman I've got `nexmo-developer` running ok in docker now. Is there an easy way to get it to render an OAS 3 definition which isn't yet published to the `api-specification` repo?

ChrisGuzman:
@MikeRalphson try it out on this branch. https://github.com/Nexmo/nexmo-developer/pull/673

For future reference you can do this on your own by following these steps:

Edit `open_api_constraint.rb` to add a new name e.g. `reporting`, create `_open_api/definitions/reporting.yml` then visit `localhost:3000/api/reporting`

MikeRalphson:
Hi @ChrisGuzman - I've split the message response into MO/MT models differentiated by one-item enums to the `type` property. This allows the use of `oneOf`.

I've locally tested adding `application/xml` to the parser/renderer (https://github.com/Nexmo/nexmo-developer/pull/677) and it seems to work fine.

What doesn't seem to work in the Nexmo renderer is having a `oneOf` inside the `items` for an `array` - as in the `messages` model. This displays ok in swagger-editor / swagger-ui.

PR #33, opened 2018-04-17
Add application OpenAPI definition

MikeRalphson:
First cut of application API OpenAPI definition.

* May need some additional `minLength` / `maxLength` properties
* I've described the properties from HAL which appear in the responses (e.g. `_embedded` and `_links`), let me know if anything else needs documenting such as 'curies' or taking descriptions from the HAL specification
* `_links` don't seem to appear in xml responses, but I've re-used the same schema for simplicity

### DIFF
--- a/definitions/application.yml
+++ b/definitions/application.yml
@@ -1,0 +1,397 @@
+openapi: 3.0.0
+info:
+  title: Nexmo Application API
+  version: 1.0.0
+  description: >-
+    A Nexmo application contains the security and configuration information you
+    need to connect to Nexmo endpoints and easily use our products.
+  contact:
+    name: Nexmo.com
+    email: devrel@nexmo.com
+    url: 'https://developer.nexmo.com/'
+    x-twitter: Nexmo
+  termsOfService: 'https://www.nexmo.com/terms-of-use'
+  license:
+    name: The MIT License (MIT)
+    url: 'https://opensource.org/licenses/MIT'
+  x-logo:
+    url: 'https://twitter.com/Nexmo/profile_image?size=original'
+  x-apiClientRegistration: 'https://dashboard.nexmo.com/sign-up'
+servers:
+  - url: 'https://api.nexmo.com/v1/applications'
+externalDocs:
+  url: 'https://developer.nexmo.com/api/developer/application'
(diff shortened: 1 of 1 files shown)


ChrisGuzman:
Looks good to me. I'm unable to get the XML responses to to render in developer.nexmo.com but I think that's a bug with the parser. Are there any specific questions I can help answer @MikeRalphson ? Feel free to comment on specific lines where you think I should pay close attention to.

definitions/application.yml
```diff
@@ -0,0 +1,397 @@
...
+          application/x-www-form-urlencoded:
+            schema:
+              type: object
+              required:
+                - name
+                - type
+                - answer_url
+                - event_url
+              properties:
+                name:
```
MikeRalphson: Are there any minLength/maxLength constraints for the `name` property?

definitions/application.yml
```diff
@@ -0,0 +1,397 @@
...
+                  type: string
+                  example: POST
+      responses:
+        '201':
+          description: Created
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/application'
+            application/xml:
```
MikeRalphson: This is possibly the reason the xml responses are not rendering. This is the media-type actually returned by the live API, but the renderer may be looking for `text/xml` ?

ChrisGuzman: Yup the renderer was looking for text/xml. In your opinion what's the best solution? Change the OAS doc to say `text/xml` or change the parser to understand `application/xml `?

MikeRalphson: I'd definitely fix the parser/renderer to understand both `text/xml` and `application/xml`. The OAS document should reflect what the API is really sending - that allows users to take your OAS document and use it with other tools.

mheap:
> _links don't seem to appear in xml responses, but I've re-used the same schema for simplicity

@MikeRalphson It will mean more duplication, but I think that the OAS should specify what is returned by the production API rather than what's easy for us to maintain. Experiencing pain documenting APIs is a good indicator that we should fix the production API

MikeRalphson:
@mheap - understood, Will amend.

MikeRalphson:
@mheap response models split into base and xml variants where needed.

PR #36, opened 2018-05-01
Add developer-account OpenAPI definition

MikeRalphson:
First cut of developer-account OpenAPI definition.

Don't think there's too much to add here, unless the API accepts `multipart/form-data` (markdown doesn't specify) or if `POST` requestBody properties can also appear as `query` parameters?


lornajane:
This is replaced by the new combined account definition spec (but I referred to this one when I made it, thanks @MikeRalphson !)

PR #37, opened 2018-05-02
Add developer-numbers OpenAPI definition

MikeRalphson:
First cut of developer-numbers API definition.

The markdown seems to contain a couple of cut-and-paste errors, in that `count` and `numbers` are given as descriptive properties (twice) for the `response` object. I'm assuming the JSON/XML examples are correct.

"An paginated" has been corrected to "A paginated" in four descriptions, as per PR https://github.com/Nexmo/nexmo-developer/pull/687

I'm assuming the description for `msisdn` in "Update a number" should read "One of your inbound numbers" or similar, and not "An available inbound virtual number"?

There are no examples of failure cases for the `POST` requests. It would be useful to know what the expected / returned HTTP status codes and `error-code` / `error-code-label` values are.

### DIFF
--- a/definitions/developer-numbers.yml
+++ b/definitions/developer-numbers.yml
@@ -1,0 +1,362 @@
+openapi: 3.0.0
+info:
+  title: Nexmo Developer Numbers API
+  version: 1.0.0
+  description: >-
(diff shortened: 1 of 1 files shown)


marklewin:
This is a great start and better than what we currently have. Merging.

PR #40, opened 2018-06-29
update stich documantation

JonathanFarrow117:


definitions/stitch.yml
```diff
@@ -150,11 +118,7 @@ paths:
           $ref: '#/components/responses/ClientError'
         '404':
           $ref: '#/components/responses/NotFound'
-      requestBody:
```
ChrisGuzman: Why are we getting rid of this?

definitions/stitch.yml
```diff
@@ -193,11 +157,7 @@ paths:
...
```
ChrisGuzman: Why are we getting rid of this?

definitions/stitch.yml
JonathanFarrow117: Because GETS cannot have request body s

definitions/stitch.yml
JonathanFarrow117: Delete can not have request bodys

definitions/stitch.yml
ChrisGuzman: Ok thanks!

definitions/stitch.yml
ChrisGuzman: ok thanks!

PR #41, opened 2018-07-04
Support requested clarification of message type.

tbedford:
My understanding is:

"If your message contains characters that can be encoded according to the GSM Standard and Extended tables then you can set the `type` to `text`. If your message contains characters outside this range, then you will need to set the `type` to `unicode`."

However, my understanding may not be correct, so I would be really grateful if someone could confirm this one way or the other.

### DIFF
--- a/definitions/sms.yml
+++ b/definitions/sms.yml
@@ -124,7 +124,7 @@ components:
           pattern: '\d{7,15}'
           example: 447700900000
         text:
-          description: The body of the message being sent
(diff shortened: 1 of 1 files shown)


mheap:
@sammachin Can you confirm if this is correct?

> "If your message contains characters that can be encoded according to the GSM Standard and Extended tables then you can set the type to text. If your message contains characters outside this range, then you will need to set the type to unicode."

sammachin:
Yes thats sounds correct as per my understanding of the API, GSM has 3 modes: 
GSM 7Bit (with local extension tables)
UTF-8
UTF-16
But we just map that to 2 options, `text` and then figure out which GSM7bit charset to use or `unicode` and use UTF-16

PR #44, opened 2018-08-13
Add in property that was in OAS doc in NDP

ChrisGuzman:
I deleted the OAS doc on NDP https://github.com/Nexmo/nexmo-developer/pull/881

and this was the only change that was on NDP but not in this doc.


ChrisGuzman:
Reviewed on my own. Merging in now but won't cut a release since it's a minor change and can be bundled with the next release

PR #47, opened 2018-10-17
fix: incorrect reference to Message change to higher level SMS schema

peterjgrainger:
This issue was raised in the documentation repo but it applies here.

https://github.com/Nexmo/nexmo-developer/issues/1016

The JSON response was referencing the schema Message where it should return the whole SMS schema.  

The response stated that one message would be returned where there will be an array of messages returned with a message count.

Changed the reference to SMS instead of message and bumped the minor version

XML reference is fine.

### DIFF
--- a/definitions/sms.yml
+++ b/definitions/sms.yml
@@ -62,7 +62,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/Message'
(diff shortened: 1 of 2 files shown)


mheap:
I've just checked locally and this looks great. Thanks for the contribution @peterjgrainger 🙌

PR #49, opened 2018-10-26
Add link to error status codes in Troubleshooting Failed SMS concept doc

marklewin:
Responding to urgent request from support to document these. The new Troubleshooting Failed SMS concept doc is PR1162 on `nexmo-developer`.

@mheap Would appreciate your help getting this into NDP.

### DIFF
--- a/definitions/sms.yml
+++ b/definitions/sms.yml
@@ -321,7 +321,7 @@ components:
             name: messageId
         status:
           type: string
-          description: The status of the message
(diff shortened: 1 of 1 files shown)


mheap:
The "correct' way to handle this is to build a new gem and update NDP to depend on it. In the interests of speed (and given our move to submodules) I've copied the YAML file in to the NDP repo instead and it will load it directly. For any future changes, pleas edit the file in NDP https://github.com/Nexmo/nexmo-developer/blob/master/_open_api/definitions/sms.yml

PR #54, opened 2018-11-23
SMS OAS validating

mheap:
With these changes, our OAS validates fine in the Swagger editor


mheap:
@lornajane There was one failure as the `info` section was missing contact details. Other than that, the spec is validating for me

```
xray ~n/api-specification/definitions
± speccy lint -s openapi-tags -s operation-tags -s default-and-example-are-redundant sms.yml                              validate-sms
Specification is valid, with 0 lint errors
xray ~n/api-specification/definitions
± speccy --version                                                                                                        validate-sms
0.8.5
```

PR #65, opened 2018-12-14
Combined Account API definition

lornajane:
As we move to making our Account API a more coherent experience, this spec aims to describe both the original "developer" account API (balance enquiry, top up and some configuration) as well as the secret management features. It's a challenge with different auth methods and even different root URLs.

Note that there are some rendering bugs that are holding this up, particularly https://github.com/Nexmo/oas_parser/issues/16


mheap:
I believe all of the blocking issues for this PR have now been resolved

lornajane:
@mheap OK so no longer blocked - do you want to review it (I feel like you've unblocked it rather looked at it) before one of us merges?  Should subaccounts be in here also?

mheap:
@lornajane I can review - wasn't sure if there was any more work to be done.

I'm happy to review + merge this, and we can add subaccounts in a followup PR as they're on EA only at the moment

definitions/account.yml
```diff
@@ -0,0 +1,551 @@
...
+                properties:
+                  type:
+                    $ref: '#/components/schemas/errorType'
+                  title:
+                    $ref: '#/components/schemas/errorTitle'
+                  detail:
+                    $ref: '#/components/schemas/errorDetail'
+                  instance:
+                    $ref: '#/components/schemas/errorInstance'
+        '404':
```
mheap: We can use our shared 404 definition here I think. See https://github.com/Nexmo/api-specification/blob/master/definitions/application.v2.yml#L298-L299

definitions/account.yml
```diff
@@ -0,0 +1,551 @@
...
+              type: string
+              description: The URI for this resource
+    errorType:
+      type: string
+      description: A link to documentation on the type of error that occurred
+      example: >-
+        https://developer.nexmo.com/api-errors/account/secret-management#validation
+    errorTitle:
+      type: string
+      description: Summary description of the error
```
mheap: Instead of generic error descriptions we should show the exact errors that can be returned. This may involve using `oneOf` if there can be multiple errors for a single HTTP code

definitions/account.yml
```diff
@@ -0,0 +1,551 @@
...
+      tags:
+        - Balance
+      responses:
+        '200':
+          description: The current balance of your account
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/accountBalance'
+              example: {"value":10.17127500,"autoReload":false}
```
mheap: We don't use `example` in any other specs. Do you think that it's useful enough to keep with the risk that it could become out of date? We generate examples based on the `schema` in NDP

definitions/account.yml
```diff
@@ -0,0 +1,551 @@
...
+        return a 200 OK response before the setting is saved.
+      parameters:
+        - $ref: '#/components/parameters/api_key_for_auth'
+        - $ref: '#/components/parameters/api_secret_for_auth'
+      tags:
+        - Configuration
+      requestBody:
+        required: false
+        content:
+          application/x-www-form-urlencoded:
```
mheap: Does this endpoint only support formencoded, or does it support JSON too?

definitions/account.yml
```diff
@@ -0,0 +1,551 @@
...
+      responses:
+        '200':
+          description: OK. Settings were updated if supplied, the details of the current settings are included in the response.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/accountSettings'
+              example: >-
+                {
+                  "mo-callback-url": "https:\/\/fabulous.example.com\/incoming\/sms",
```
mheap: We use the following URLs in the building blocks:

* `/webhooks/inbound-sms`
* `/webhooks/delivery-receipt`

definitions/account.yml
lornajane: Yes, I think all our specs should have examples in them and be designed for use outside of NDP as well as there.

definitions/account.yml
lornajane: I think only form was documented but I'll poke the endpoint to be certain, good spot

definitions/account.yml
lornajane: I want to move that file so I'll defer this change rather than blocking this PR - it would be nice to revisit all the specs for where the shared definitions can be reused and be consistent.

definitions/account.yml
lornajane: I can't find any docs or persuade it to accept JSON by trying! So I'm going with no, it doesn't support JSON.

definitions/account.yml
lornajane: Ah OK. In that case I think all the error parts should be per-path-endpoint, and not `$ref` into `components`. I thought it was enough to describe the error format used but the specifics are different for the different endpoints.

definitions/account.yml
lornajane: Updated

lornajane:
@mheap thanks for the review, bunch of things looking better now :)  If you want the specific errors rather than just field descriptions then I have more work to do but I think that's the only thing outstanding now!

definitions/account.yml
mheap: For most errors we use a common format, so `$ref` works well (e.g. you can get a `404` on `GET`, `PUT` and `DELETE` and it should show the same error in all). For single use errors I agree that it should be per-path-endpoint

mheap:
Yes please on specific errors

I also need to double check response serialisation in NDP for formencoded, but once those two things are done I think we're good to merge

PR #67, opened 2018-12-14
Use Verify OAS

marklewin:
Modified Verify OAS to incorporate any "good bits" from the existing markdown-based API ref in NDP. Added XML responses, ensured that everything rendered correctly.


definitions/verify.yml
```diff
@@ -693,45 +698,45 @@ components:
...
-              - value: null
-                description: >-
-                  Cancel: You must wait at least 30s after sending a Verify
-                  Request before cancelling, or Verify has made too many
-                  attempts to redeliver a PIN for this request; you have to wait
-                  for the workflow to complete. Also, you cannot initiate a new
-                  Verify Request until this one expires. Trigger_next_event: All
-                  the attempts to deliver the PIN for this request have been
-                  completed and there are no more events to skip to.
+          # x-ms-enum:
```
mheap: Can we show this information in another format like we do above in tables?

definitions/verify.yml
```diff
@@ -5,32 +5,8 @@ info:
...
   version: 1.0.0
   description: >-
-    Nexmo's Verify API allows you to send a PIN by SMS and phone to prove a user
-    can be contacted at a specific phone number.
-
-
-    This is useful for:
-
-
-    * Spam protection - prevent spammers from mass-creating new accounts (etc.)
```
mheap: The use cases may be useful in the description at the top

definitions/verify.yml
```diff
@@ -693,45 +698,45 @@ components:
         status:
           type: string
           description: >-
-            The Verify Control Response code that explains how your request
+            The Verify control response code that explains how your request
             proceeded.
           example: '0'
           enum:
             - '0'
             - '19'
```
mheap: What does error code 19 mean?

definitions/verify.yml
```diff
@@ -693,45 +698,45 @@ components:
...
           type: string
           description: The `cmd` you sent in the request.
           example: cancel
         error_text:
           type: string
-          description: 'If status is not 0, this explains the error encountered.'
+          description: If the `status` is non-zero, this explains the error encountered.
           example: error
   securitySchemes:
     apiKey:
```
mheap: This is a required field

definitions/verify.yml
```diff
@@ -693,45 +698,45 @@ components:
...
       type: apiKey
       name: api_key
       in: query
       description: >-
-        You can find your API key in your [account
-        overview](https://dashboard.nexmo.com/account-overview)
+        You can find your API key in your Nexmo account [developer
+        dashboard](https://dashboard.nexmo.com/)
     apiSecret:
```
mheap: This is a required field

definitions/verify.yml
```diff
@@ -228,36 +205,25 @@ paths:
...
-            *   `pin_expiry` = 360 seconds, so `next_event_wait` = 120 seconds -
-            all three attempts have the same PIN.
-
-            *   `pin_expiry` = 240 seconds, so `next_event_wait` = 120 seconds -
-            1st and 2nd attempts have the same PIN, third attempt has a
-            different PIN.
-
-            *   `pin_expiry` = 120 (or 200 or 400 seconds) - each attempt has a
-            different PIN.
+            How log the generated verification code is valid for, in seconds.
```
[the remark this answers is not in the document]
marklewin: Instead of describing this inline, I link to the new Changing the Default Event Timings concept doc.

definitions/verify.yml
```diff
@@ -338,18 +303,21 @@ paths:
...
+                $ref: '#/components/schemas/checkResponse'                
   '/search/{format}':
     get:
       description: >-
-        1. Send a Verify Search request containing the `request_id`s of the
-        Verify requests to search for.
+        To check the status of past or current verification requests:
+
+
+        * Send a Verify Search request containing the `request_id`s of the
```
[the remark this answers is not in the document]
marklewin: No, and I agree that these should be numbered lists. Changed.

definitions/verify.yml
marklewin: Added a table for this information.

definitions/verify.yml
marklewin: Added back! I wanted people to look at the Overview for this but agree that it's probably useful here too.

definitions/verify.yml
marklewin: It's a control request failure. I've tested this and can confirm and so I've put the information in the table of response codes.

definitions/verify.yml
marklewin: This is a response field, not a request field?

definitions/verify.yml
marklewin: Good spot! Changed.

definitions/verify.yml
mheap: `apiKey`? It's a request `query` parameter I believe. That's how it shows once rendered

definitions/verify.yml
lornajane: I've been adding links to overview pages into the OAS description fields in an attempt to help tourists find the "real" documentation. Might be a good tactic here too.

definitions/verify.yml
```diff
@@ -479,66 +454,86 @@ components:
...
+              #     more than 3 times, it is set to FAILED and you cannot check it
+              #     again.
+              # - value: '18'
+              #   description: >-
+              #     You added more than the maximum of 10 `request_id`s to your
+              #     request.
+              # - value: '19'
+              #   description: No more events are left to execute for the request
+              # - value: '101'
+              #   description: There are no matching Verify requests.
```
lornajane: Don't be afraid to delete rather than comment out once you're done and don't need to refer to this bit - that's what source control is for.

definitions/verify.yml
lornajane: We've got something weird going on with authentication where we don't fully describe it in the spec and some of it is in the render. I didn't find a good way to describe this sort of Auth in OpenAPI and have instead added the parameters to each request as required query params. I think this makes it clearer to users what they should do and also means it pops up in any other tooling that is used. It's fine to leave it old-style for now but if you're interested, the Account API spec is where I was working on this: https://github.com/Nexmo/api-specification/pull/65/files#diff-59e397c2ded341a84dccb685f721617bR20

mheap: This spec does the same for auth. They're defined as security schemes, but used as query parameters as a `ref`

PR #70, opened 2019-01-03
Read callback deprecated for Viber

tbedford:
## Description 

Read callback was deprecated on 19th Dec 2018 by Viber to reduce load.


hughhopkins:
@tbedford thanks. Can we add that WhatsApp has read status please

PR #75, opened 2019-02-06
Update verify.yml

paperbag-zz:
The api_key & api_secret does not have a "required" field underneath the parameters for Verify specs


paperbag-zz:
Re-doing this, so it doesn't fail!

PR #85, opened 2019-03-15
Add new VBC extension endpoint

marklewin:


marklewin:
Removing as there is an embargo against any VBC stuff being advertised before Enterprise Connect on Monday 18th March.

PR #86, opened 2019-03-15
Add VBC extension endpoint

marklewin:
Add new VBC extension endpoint in `createCall`

Please **DO NOT MERGE** because this stuff needs to be kept secret until Enterprise Connect on Monday 18th March.


marklewin:
This can be seen here: https://nexmo-developer-vbc.herokuapp.com/api/voice#createCall

PR #88, opened 2019-03-20
Mention that POST requests are also accepted

marklewin:
# Description

The Number Insight APIs accept `POST` as well as `GET` requests. We don't want to advertise this (the NI API is read-only, making `GET` the most appropriate verb for requests) but we do need to document it.

### DIFF
--- a/definitions/number-insight.yml
+++ b/definitions/number-insight.yml
@@ -28,6 +28,10 @@ paths:
     get:
       operationId: getNumberInsightBasic
       summary: Basic Number Insight
+      description: |
+        Provides [basic number insight](/number-insight/overview#basic-standard-and-advanced-apis) information about a number.
+
+        Note that this endpoint also supports [POST] requests.
       parameters:
         - $ref: '#/components/parameters/number'
         - $ref: '#/components/parameters/country'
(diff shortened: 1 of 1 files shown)


lornajane:
Can you bump the version number on the spec as well please? I tried to add a pull request template to prompt people so feedback on whether that was visible/working/clear would also be welcome!

definitions/number-insight.yml
```diff
@@ -28,6 +28,10 @@ paths:
     get:
       operationId: getNumberInsightBasic
       summary: Basic Number Insight
+      description: |
+        Provides [basic number insight](/number-insight/overview#basic-standard-and-advanced-apis) information about a number.
+
+        Note that this endpoint also supports [POST] requests.
```
lornajane: Can this be `POST` with backticks rather than square brackets? I think it makes it clearer (and the same in the other occurrences)

marklewin:
I didn't notice that! (Which, of course, doesn't mean that it wasn't there.) I'll pay more attention next time, but great idea :)

> Can you bump the version number on the spec as well please? I tried to add a pull request template to prompt people so feedback on whether that was visible/working/clear would also be welcome!

JcChimjj:
＃88

PR #90, opened 2019-03-21
Verify API request limit

marklewin:
# Description

See [DOCS-344](https://nexmoinc.atlassian.net/browse/DOCS-344). Describe limit of one verify request per second.

**Notes**:

* Used this as an opportunity to remove tags, which don't add anything here and make the headings for each endpoint look HUGE.
* I'll also need to add this to the main docs under [PR 1582](https://github.com/Nexmo/nexmo-developer/pull/1582).

# Checklist


tbedford:
The first file in the file change list is to do with Number Insight API but the description only mentions Verify. Is that additional file intentional? 

Is there a review app or something I can look at?

lornajane:
@marklewin can we clean up the commit history here and remove the mistaken commits and their reverts? Let me know if I should do that, I'm happy to. Then we can review

marklewin:
In answer to your questions @tbedford :

> The first file in the file change list is to do with Number Insight API but the description only mentions Verify. Is that additional file intentional?

My bad - change reverted and commit history rebased as per @lornajane's request.
 
> Is there a review app or something I can look at?

Yes, it's at [PR 1582](https://github.com/Nexmo/nexmo-developer/pull/1582).

PR #91, opened 2019-03-22
Move the common files to a subdirectory of definitions

lornajane:
# Description

Moving the dependencies to a subdirectory so it's more clear they belong with the definitions files. Renaming to "common" as I think it's more clear that things there are common standards.

# Checklist


mheap:
This PR has made me realise that we have a bug in our existing speccy checks - it only checks top level `.yml` files. It should be `/**/*.yml`

https://github.com/Nexmo/api-specification/blob/master/bin/validate-specs.sh#L16

If we move to a `common` folder for this, we'll need to exclude this folder from speccy linting, or ensure that each reference is also a valid OAS. I'm leaning towards the former - any thoughts?

lornajane:
Agreed. We should recurse but omit  a `common` directory

mheap:
Merging this with the linting bug, and will omit `common` when I look at recursing the linting check

PR #94, opened 2019-03-26
Add `vbc` capability

marklewin:
# Description

Documenting the `vbc` capability, which is required for zero-rated calls to the VBC number programmability service. Also fixed indenting.

# Checklist

### DIFF
--- a/definitions/application.v2.yml
+++ b/definitions/application.v2.yml
@@ -147,6 +147,10 @@ paths:
                                   example: POST
                                   enum:
                                     - POST
+                    vbc:
+                      type: object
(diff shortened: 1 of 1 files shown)


definitions/application.v2.yml
```diff
@@ -147,6 +147,10 @@ paths:
                                   example: POST
                                   enum:
                                     - POST
+                    vbc:
+                      type: object
+                      description: "Specify `vbc` capability to enable zero-rated calls for VBC number programmability service applications."
```
mheap: I'd add a note here explicitly calling out that `vbc` should be an empty object

PR #101, opened 2019-04-08
Tidy POST parameters in Numbers OAS

marklewin:
# Description

Pass all API key and secret params as query parameters ("list owned" also supports API key and secret in the path, but we shouldn't advertise this as it's not consistent across all endpoints).

# Checklist

### DIFF
--- a/definitions/numbers.yml
+++ b/definitions/numbers.yml
@@ -1,7 +1,7 @@
 openapi: 3.0.0
 info:
   title: Numbers API
-  version: 1.0.3
+  version: 1.0.4
   description: >-
     The Numbers API enables you to manage your existing numbers and buy new virtual numbers for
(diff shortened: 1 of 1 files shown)


definitions/numbers.yml
```diff
@@ -90,18 +89,17 @@ paths:
...
-          description: An available inbound virtual number
-          schema:
-            type: string
-          example: '447700900000'
+      description: Request to purchase a specific inbound number.   
+      requestBody:
+        description: Number details
+        required: true
+        content:
+          application/json:
```
mheap: I believe that these endpoints only support `application/x-www-form-urlencoded` content

marklewin: Yes, my silly mistake. Fixed.

PR #105, opened 2019-04-12
Add Voice V2 OAS

mheap:
It is a `beta`, with only one endpoint (`POST /calls`) supported. The
only difference to v1 is that this request returns a 202 rather than
a 201 containing a call ID.

This change is required as a call ID is not generated until a call is
placed, which can be at an indeterminate point in the future for IP legs


mheap:
Got a 👍 from Jamie on the VAPI team

PR #121, opened 2019-05-07
Advise against use of Sync method for NI Advanced

marklewin:
# Description

See [DEVX-148](https://nexmoinc.atlassian.net/browse/DEVX-1048).

TL/DR: Using NI Advanced synchronously leads to timeouts, so we should de-empahasize this capability. I have a number of changes to overview, concepts and tutorials that I'll push to NDP along with this change once approved.

# Checklist

### DIFF
--- a/definitions/number-insight.yml
+++ b/definitions/number-insight.yml
@@ -4,7 +4,7 @@ servers:
   - url: 'https://api.nexmo.com/ni'
 info:
   title: Number Insight API
-  version: 1.0.2
+  version: 1.0.3
   description: >-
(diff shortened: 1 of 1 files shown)


definitions/number-insight.yml
```diff
@@ -138,6 +113,33 @@ paths:
...
+  '/advanced/{format}':
+    parameters:
+      - $ref: '#/components/parameters/format'
+    get:
+      operationId: getNumberInsightAdvanced
+      summary: Advanced Number Insight (sync)
+      description: |
+        Provides [advanced number insight](/number-insight/overview#basic-standard-and-advanced-apis) information about a number synchronously, in the same way that the basic and standard endpoints do.
+
+        Nexmo recommends accessing the Advanced API **asynchronously**, to avoid timeouts.
```
lornajane: Could we make this even clearer by saying to use the `/advanced/async` endpoint specifically?

marklewin: Good idea! Added.

PR #123, opened 2019-05-09
Bring Verify spec in line with live API

lornajane:
# Description

The spec for Verify hasn't been kept in step with changes to the API. Also since the API has some unusual features, it's quite tricky to document. The main changes are:

- document new `events[]` parameter in result of `/search` endpoint
- try to tidy up in general, warn about `require_type` which will cause an error response to be returned for most users if they try to use it, remove commented bits of spec.
- add error responses (our errors are have status 200, and I'm really unclear which status can be returned for which endpoint but this is my first stab at this problem).

# Checklist

### DIFF
--- a/definitions/verify.yml
+++ b/definitions/verify.yml
@@ -597,6 +597,23 @@ components:
               ip_address:
                 type: string
                 example: 123.0.0.255
+        events:
+          type: array
+          xml:
+            wrapped: true
+          description: The events that have taken place to verify this number, and their unique identifiers.
+          items:
+            type: object
+            xml:
+              name: event
+            properties:
+              type:
+                type: string
+                enum:
+                  - tts
+                  - sms
+              id:
+                type: string
         error_text:
           type: string
           description: If `status` is not `SUCCESS`, this message explains the issue encountered.
--- a/definitions/verify.yml
+++ b/definitions/verify.yml
@@ -3,7 +3,7 @@ servers:
   - url: 'https://api.nexmo.com/verify'
 info:
   title: Nexmo Verify API
-  version: 1.0.2
+  version: 1.0.3
   description: >-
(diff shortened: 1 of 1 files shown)


definitions/verify.yml
```diff
@@ -220,7 +221,7 @@ paths:
             How log the generated verification code is valid for, in seconds.
             When you specify both `pin_expiry` and `next_event_wait` then `pin_expiry` must be 
             an integer multiple of `next_event_wait`. Otherwise, `pin_expiry`
-            is set to equal next_event_wait. See [changing the default event timings](/verify/guides/changing-default-timings).
+            is set to equal next_event_wait. See [changing the default event timings](https://developer.nexmo.com/verify/guides/changing-default-timings).
```
marklewin:
```suggestion
            is set to equal `next_event_wait`. See [changing the default event timings](https://developer.nexmo.com/verify/guides/changing-default-timings).
```

definitions/verify.yml
```diff
@@ -678,10 +679,56 @@ components:
...
+          enum:
+            - IN PROGRESS
+            - FAILED
+            - EXPIRED
+            - CANCELLED
+            - '101'
+          description: |
+            Code | Description
+            -- | --
+            IN PROGRESS | The search is still in progress.
```
marklewin: Can we use code font here?

marklewin: You can use them in the `description`:

```
          description: |
            Code | Description
            -- | --
            `IN PROGRESS` | The search is still in progress.
            `SUCCESS` | Your user entered a correct verification code.
            `FAILED` | Your user entered an incorrect code more than three times.
            `EXPIRED` | Your user did not enter a code before the `pin_expiry` time elapsed.
            `CANCELLED` | The verification process was cancelled by a Verify control request.
            `101` | You supplied an invalid `request_id`.
```

PR #130, opened 2019-05-29
VNI-184 :: [Custom Workflow] Update OpenSpec API

Dbasdeo:
# Description

What does this pull request achieve? Why do we need it?

# Checklist

### DIFF
--- a/definitions/verify.yml
+++ b/definitions/verify.yml
@@ -3,7 +3,7 @@ servers:
   - url: 'https://api.nexmo.com/verify'
 info:
   title: Nexmo Verify API
-  version: 1.0.2
+  version: 1.0.3
   description: >-
     The Verify API helps you to implement 2FA (two-factor authentication) in your applications.
     This is useful for:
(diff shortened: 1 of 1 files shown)


lornajane:
Are these fields available to all users? It was my understanding that only certain customers with particular pricing plans would have access - can you clarify?

Dbasdeo:
The pin_code parameter is not available to all users however the workflow_id is. To let people know that pin_code is not available to everyone, in my description, I added: "You must contact 
            [support@nexmo.com](mailto:support@nexmo.com) to enable this 
            feature." 
I copied this line from another parameter that is also not available to all users.

lornajane:
Thanks for clarifying! I am starting by creating the workflow docs since we need to be able to link to these and they will be part of our public documentation. Additional docs for "secret" features can follow but I think all not-generally-available fields should be removed from the spec since users aren't able to tell for themselves what pricing model they have and this could cause confusion.

lornajane:
This PR is superseded by #131 which includes only the features available to all users (we're creating custom docs for the extra features) and also rebases onto the recent changes to this spec.

PR #131, opened 2019-06-03
Add Verify workflow fields to spec

lornajane:
# Description

We have added some new fields to the Verify API. These allow the user to specify which workflow to use for each verify request.

This PR also removes the `require_type` field which is not available to all users, and adds the status response for errors involving use of the `pin_code` field for accounts that don't support that functionality.

# Checklist

### DIFF
--- a/definitions/verify.yml
+++ b/definitions/verify.yml
@@ -3,7 +3,7 @@ servers:
   - url: 'https://api.nexmo.com/verify'
 info:
   title: Nexmo Verify API
-  version: 1.0.3
+  version: 1.0.4
   description: >-
     The Verify API helps you to implement 2FA (two-factor authentication) in your applications.
     This is useful for:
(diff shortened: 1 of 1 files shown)


lornajane:
Did we approve but not merge for a reason? I'll add this to the NDP PR when it's in master

definitions/verify.yml
```diff
@@ -239,7 +225,26 @@ paths:
...
             default: 300
-          example: 300
+          example: 120
+        - name: workflow_id
+          required: false
+          in: query
+          description: >-
+            Selects the predefined sequence of SMS and TTS (Text To Speech)
+            actions to use in order to convey the PIN to your user. For example,
+            an id of 1 identifies the workflow SMS - TTS - TTS. For a list of
```
mheap: The existing SMS-TTS-TTS docs are a mistake. Have we confirmed that workflow 1 is SMS-TTS-TTS and that it is the new default? If so it's a breaking change

lornajane: I checked it after we talked the other day and by default I get SMS -> TTS -> TTS so the docs are correct and worfklow_id 1 is the correct default and reflects the current behaviour.

PR #140, opened 2019-07-10
Update VAPI PUT request spec to include inline NCCO for transfer action

benhayehudi:
# Description

There has been a change implemented for a `PUT` request for in progress calls, wherein an NCCO object can now be provided instead of a url path for the `transfer` action. For example:

```
{  
      "action":"transfer",
      "destination":{  
         "type":"ncco",
         "ncco":[  
            {  
               "action":"talk",
               "text":"hello world"
            }
         ]
      }
   }
```

# Checklist

### DIFF
--- a/definitions/voice.yml
+++ b/definitions/voice.yml
@@ -367,6 +367,8 @@ paths:
                         type: string
                     ncco:
                       type: array
+                      items:
+                        type: string
                       example: '[{"action": "talk", "text": "hello world"}]'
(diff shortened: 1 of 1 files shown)


benhayehudi:
Thank you, @marklewin for the review, particularly the `x-nexmo-developer-collection-description-shown` addition. It would be great to sit together next time when we're in person and I could pick your brain more on writing excellent API reference docs!

PR #144, opened 2019-07-17
Update PULL_REQUEST_TEMPLATE.md

mheap:
# Description

Updates the PR template to add a suggested format for PR descriptions as these are now used to auto-tag releases e.g. https://github.com/Nexmo/api-specification/releases/tag/voice-1.2.0

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,6 +1,20 @@
 # Description
 
-What does this pull request achieve? Why do we need it?
+<!--
+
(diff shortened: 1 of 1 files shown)


lornajane:
What happens if people open the PR without adding something in the Description section? We see a lot of these and haven't been strict with it in the past

mheap:
We can edit the description once the PR has been opened. If the PR is merged without it being opened it will tag a release with that message which will need updating.

I'd like to think that this is a human problem to solve, but I'd love to hear any technology solutions that are available

lornajane:
I think my solution would be to just close PRs if people can't open them properly so probably editing is more constructive!!

PR #147, opened 2019-07-17
Refactor auth to make api_secret a required param

lornajane:
# Description

Since our APIs don't use a standard authentication scheme, I have been looking for a way to explain the api key and secret approach to automated tools such as mock servers, Postman, etc. This change removes the suggestion of two acceptable security schemes, and adds the `api_secret` as a required parameter to every endpoint.

# Fixes

Make auth process clearer especially for OpenAPI parsing tools

# Checklist

**err, not incremented because this isn't the only open PR for this spec, proceed with caution**

### DIFF
--- a/definitions/verify.yml
+++ b/definitions/verify.yml
@@ -29,7 +29,6 @@ externalDocs:
   url: 'https://developer.nexmo.com/verify'
 security:
   - apiKey: []
-    apiSecret: []
 paths:
   '/{format}':
     get:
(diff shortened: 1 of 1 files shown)


judy2k:
Hello! If `api_secret` is required for all the calls, doesn't that make `api_key` required?

lornajane:
@judy2k good question! The API key is required because the security is set to api_key. But we previously had a weird setup that sort of said: you can use either of these security schemes, one requires `api_key` and one requires `api_secret`. So all the tools just pick one and the other field gets ignored completely. Our docs do an interesting fudge. I'm hoping this is is a more accurate reflection of reality.

marklewin:
This is great, and not something I was even aware of! Perhaps if we roll this out across the API refs we can remove some complexity in NDP?

lornajane:
@marklewin did you approve but not merge for a reason? I can merge, just wanted to check!

marklewin:
Left it open in case you wanted @mheap to weigh in, but happy to merge and doing so!

PR #155, opened 2019-08-16
Application API V1 deprecated

tbedford:
## Description

Deprecate V1 API. 

According to Brennan Topley App API v2 can now be considered GA. So, App v1 API definition has been removed, and references to App API v2 as being Beta have also been removed.

# New 

* Application API V1 is now deprecated.

## Checklist


tbedford:
The page has a small link box to switch between VERSION 1 and VERSION 2 that can be removed as V1 has been deprecated.

mheap:
We need to keep the V1 docs online, but make all links point to v2 by default and add a note on the v1 docs that the API is deprecated. We're likely to support for ~12 months

tbedford:
Closing. Probably best to simply create a new PR to label App v1 as deprecated.

PR #161, opened 2019-08-29
API reference documentation for REST Conversation Recording

benhayehudi:
# Description

# New 

* Added new `/conversations/{conversation_id}/record` endpoint for REST start/stop Conversation recording

# Checklist

### DIFF
--- a/definitions/conversation.yml
+++ b/definitions/conversation.yml
@@ -167,7 +167,7 @@ paths:
           $ref: '#/components/responses/SuccessEmptyJSON'
   '/conversations/{conversation_id}/record':
     parameters:
-      - $ref: '#/components/parameters/conversation_id/record'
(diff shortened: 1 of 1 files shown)


definitions/conversation.yml
```diff
@@ -165,6 +165,23 @@ paths:
...
+    parameters:
+      - $ref: '#/components/parameters/conversation_id'
+    put:
+      operationId: recordConversation
+      tags:
+        - conversation
+      summary: Record a conversation
+      responses:
+        '204':
+          description: '{}'
```
marklewin: Would suggest using `No content` rather than `{}` because it's clearer and that's how I recall seeing it done elsewhere? E.g. https://developer.nexmo.com/api/voice#updateCall

benhayehudi: Good catch, just changed it!

PR #162, opened 2019-08-30
Fix Modify Call API Docs

benhayehudi:
# Description

# Fixes

* Adjust description of Path Parameters to require `uuid` of the leg, not of the call

# Checklist

### DIFF
--- a/definitions/voice.yml
+++ b/definitions/voice.yml
@@ -1,7 +1,7 @@
 ---
 openapi: 3.0.0
 info:
-  version: 1.2.0
+  version: 1.2.1
   title: Voice API
   description: The Voice API lets you create outbound calls, control in-progress calls
     and get information about historical calls. More information about the Voice API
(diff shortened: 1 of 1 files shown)


definitions/voice.yml
```diff
@@ -391,7 +391,7 @@ paths:
       schema:
         type: string
       required: true
-      description: UUID of the Call
+      description: UUID of the Leg
```
mheap: There are more instances in this typo to change. Everything that accepts an ID is a leg operation

benhayehudi: 👍

definitions/voice.yml
```diff
@@ -623,7 +623,7 @@ components:
     uuid:
       type: string
       format: uuid
-      title: The UUID of the Call
+      title: The UUID of the Leg
```
mheap: Should this be `Call Leg` as above?

benhayehudi: 👍

PR #163, opened 2019-09-02
Update nexmo-oas-renderer.

fabianrbz:
# Description
Bump `nexmo-oas-renderer`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,4 @@
 source 'https://rubygems.org'
 
 gem 'rake'
-gem 'nexmo-oas-renderer', '~> 0.3.2'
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -1,4 +1,4 @@
 source 'https://rubygems.org'
 
 gem 'rake'
-gem 'nexmo-oas-renderer', '~> 0.3.2'
+gem 'nexmo-oas-renderer'
```
mheap: I prefer having minimum versions in here as it makes it obvious what's happening if something isn't working

mheap: Ah nevermind, this is the api-specification repo, not NDP

PR #164, opened 2019-09-02
Numbers API: Add new `has_application` and `application_id` filters

lornajane:
# Description

See also https://nexmoinc.atlassian.net/browse/DEVX-890. We added some new filtering fields to the Numbers API and those features will go live in the coming days. This is the docs update to match :)

# New 

* Added new `has_application` and `application_id` filters to the Numbers API

# Checklist

### DIFF
--- a/definitions/numbers.yml
+++ b/definitions/numbers.yml
@@ -1,7 +1,7 @@
 openapi: 3.0.0
 info:
   title: Numbers API
-  version: 1.0.8
+  version: 1.0.9
   description: >-
     The Numbers API enables you to manage your existing numbers and buy new virtual numbers for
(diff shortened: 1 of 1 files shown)


lornajane:
flagged as "don't merge" because the feature isn't actually live yet ...

definitions/numbers.yml
```diff
@@ -230,6 +232,27 @@ components:
...
           - 2
       example: 1
+    has_application:
+      name: has_application
+      required: false
+      in: query
+      description: |
+        Set this optional field to `true` to restrict your results to numbers
+        associated with an application (any application). Set to `false` to
+        find all numbers _not_ associated with any application. Omit the field
```
marklewin: Alas, Volta won't render this _not_. Fabian is looking into, but perhaps use *bold* if you want to keep the emphasis for now?

PR #170, opened 2019-09-16
Update available language codes for Verify

lornajane:
# Description

Updates the list of available language codes now that Verify is using the new TTS feature.

# Checklist


lornajane:
see also https://github.com/Nexmo/nexmo-developer/pull/2035

PR #171, opened 2019-09-17
Update voice.yml

sammachin:
# Fixes

* Corrected NCCO to be an "array of objects" not an "array of strings"

# Checklist

- [x] version number incremented (in the `info` section of the spec) set to 1.2.3


lornajane:
Definitely a feature for @benhayehudi to take a look at but I've pushed the same change in a different place in the same spec, and also bumped the version number. There's a checkbox in the PR template for the version increment but we don't seem to have used the template this time and that is usually what causes us to miss the version change so I just added it while I was thinking of it!!

tbedford:
@benhayehudi can you take a look at this PR when you get a chance please? I fixed the merge conflict (hopefully) and also changed the PR description to the correct format. Thanks.

PR #172, opened 2019-09-20
Switch to basic spectral validation of built-in rules

lornajane:
# Description

Move from our existing [speccy](https://speccy.io/) to [spectral](https://github.com/stoplightio/spectral). The main reasons for the move:

* Spectral is more actively developed and supported
* Spectral has out-of-the-box validation (some of which I've turned off for now)
* It supports adding custom rulesets so we'd be able to add more opinionated checks in line with our API standards effort

### DIFF
--- a/.spectral.yml
+++ b/.spectral.yml
@@ -1,0 +1,9 @@
+extends: spectral:oas3
+rules: 
+  operation-tags: false
+  operation-description: false
+  valid-example-in-schemas: false
+  valid-schema-example-in-content: false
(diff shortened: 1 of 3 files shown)


mheap:
@lornajane I updated CI and this looks good to me. What else do we need to do to ship it?

lornajane:
I think we can ship it! There's more we could do but this puts us at parity (and more) with the rules we already had and we can build on Speccy from here. Removing the Do not merge label

PR #178, opened 2019-10-02
Remove rake

timcraft:
Removes the Rakefile and the rake gem dependency.

Rake doesn't appear to be used for anything—there aren't any rake tasks defined, and the travis validation script is using bash/node.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,3 @@
 source 'https://rubygems.org'
 
-gem 'rake'
 gem 'nexmo-oas-renderer', '~> 0.6.1'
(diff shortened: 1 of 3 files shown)


fabianrbz:
Hi @timcraft, we do have some tasks defined under `lib/tasks/` and we also use Rails' ones (you can take a look at them by running `rake -T`) so we can't remove this dependency

tbedford:
@fabianrbz - I think there are conflicts here that need resolving? Is that right that Rake is not used? I thought I saw in a comment it is used? Thanks.

fabianrbz:
@tbedford we need to fix the conflicts. Yeah, I edited the comment but somehow i ended up deleting it 🤦‍♂ . We don't use it in this repo.

timcraft:
@tbedford @fabianrbz I've rebased and fixed up the conflict in the Gemfile.

Confusing I know but `lib/tasks` doesn't exist in the repo, and `rake -T` displays nothing.

tbedford:
Thanks Tim!

PR #179, opened 2019-10-04
Added missing secret parameter

obodriagovatvonagedotcom:
When creating the subaccount, the primary account should be able to set a custom secret for this subaccount. The code supports it, bu the documentation does not reflect this.

# Description

# Checklist


mheap:
Definitely forgot to check that the version number was increased here. My bad, I'll resolve it

PR #202, opened 2019-11-07
update vbc call recording apis

bdorry:


marklewin:
Applied to `markl-new-vbc-recording-apis` branch

PR #209, opened 2019-11-12
Bdorry new vbc recording apis

bdorry:


marklewin:
Applied direct to branch

PR #220, opened 2019-11-26
Add some example responses to verify API

lornajane:
# Description

We have had a request to enable testing of the Verify API and its various responses. One way to achieve this is using prism as a local mock server which would enable clients to test against the ideal operation of the API and to reproduce error cases that would be hard to achieve in other ways.

This PR introduces many more examples to the Verify spec that can be used to test the API.

# How to use this feature

1. Install Prism https://github.com/stoplightio/prism

2. Check out this branch and start prism serving the spec: `prism mock ~/specs/verify.yml`

3. Make requests to the mock server (by default on port 4010) that is now running, e.g.:

  `curl "http://localhost:4010/json?api_key=123&api_secret=456&brand=sequin&number=447900797979"`

4. Now explore other possible responses by adding the `__example` parameter and the name of the response you'd like to get, such as:

  `curl "http://localhost:4010/json?api_key=123&api_secret=456&brand=sequin&number=447900797979&__example=concurrent"`

  `curl "http://localhost:4010/json?api_key=123&api_secret=456&brand=sequin&number=447900797979&__example=throttled"`

# What this PR does not include:

We don't know which tech stack we are targeting, it might be nice to have some instructions for how to integrate this with an existing application. For example some of the SDKs support changing the target URL which will help with testing.

We were also asked for latency, I'm investigating if prism can add a delay.

**Version is not updated, these instructions/papertrail need to be removed from this PR because the content gets used in the changelog**

# New

* improved support for examples in Verify API to aid testing

# Checklist

### DIFF
--- a/definitions/verify.yml
+++ b/definitions/verify.yml
@@ -3,7 +3,7 @@ servers:
   - url: "https://api.nexmo.com/verify"
 info:
   title: Nexmo Verify API
-  version: 1.0.10
+  version: 1.0.11
   description: >-
     The Verify API helps you to implement 2FA (two-factor authentication) in your applications.
     This is useful for:
@@ -712,10 +712,7 @@ components:
             - "9"
             - "10"
             - "15"
-            - "18"
-            - "19"
             - "20"
-            - "101"
           description: |
             Code | Text | Description
             -- | -- | --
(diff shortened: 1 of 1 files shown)


lornajane:
@roybensh this is what I have so far!

lornajane:
I think these detailed examples are useful, and there is no point in having them hanging around here! @mheap do you think we could merge these in? I have been using them to demo the self-service sandbox idea.

definitions/verify.yml
```diff
@@ -316,6 +452,63 @@ paths:
...
+              examples:
+
+                success:
+                  summary: The request was found
+                  value:
+                    request_id: abcdef0123456789abcdef0123456789
+                    account_id: abcdef01
+                    number: 447700900000
+                    sender_id: verify
+                    date_submitted: 2020-01-01 12:00:00
```
mheap: Dates need quoting to work with the renderer

https://nexmo-api-specification-pr-220.herokuapp.com/verify

```
(.//definitions/verify.yml): found unexpected ':' while scanning a plain scalar at line 466 column 46
```

lornajane:
Rebasing to pick up spectral changes which should help with the build failure

definitions/verify.yml
lornajane: good catch!

lornajane:
another rebase to pick up recent changes and bump the version number again to keep it ahead of master. We had some feedback about showing all errors under all endpoints, I'd tidied up a few already but I also cleaned up some more in a final (?) commit so I think this would be good to merge if we're comfortable.

PR #223, opened 2019-12-03
Added correct Base URL

sammachin:
The Base URL for the VIS api is wrong (https://api.vonage.com/vgis_ it looks as if it wasn't set in the OAS spec so I'm not quite sure where the renderer was picking it up from, this is the correct URL


sammachin:
found the existing incorrect URL hidden at line 1088 about 2/3 of the way through!

PR #224, opened 2019-12-09
Verify v2

lornajane:
# Description

Proposed spec for the new version of the Verify API. We need to think about how the versions should work here as I'm sure 0.0.1 makes no sense. Also this description will need to be updated to please the changelog once we are ready to merge.


definitions/verify.v2.yml
```diff
@@ -0,0 +1,427 @@
...
+        - contact
+        - company_name
+      properties:
+        events:
+          type: array
+          items:
+            oneOf:
+              - $ref: '#/components/schemas/ToMessages'
+              - $ref: '#/components/schemas/ToMessenger'
+              - $ref: '#/components/schemas/ToVoice'
```
mheap: Why is `Voice` separate here? It uses the same fields as `ToMessages`

definitions/verify.v2.yml
```diff
@@ -0,0 +1,427 @@
...
+                    title: 'The `company_name` parameter is missing or invalid'
+                    detail: 'You must supply a `company_name` parameter as a string'
+                throttled:
+                  summary: You have exceeded the API rate limit for your account
+                  value:
+                    type: 'https://www.nexmo.com/messages/Errors#Throttled'
+                    title: You have exceeded the API rate limit for your account
+                    detail: For more advice on rate limits and your current account settings, please contact support.
+
+  /requests/{request_id}/status:
```
mheap: What's the rationale for renaming this to `status` rather than `check`?

definitions/verify.v2.yml
```diff
@@ -0,0 +1,427 @@
...
+      - $ref: '#/components/parameters/request_id'
+      requestBody:
+        description: Supply a PIN to be checked
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/PinStatus'
+      responses:
+        "200":
```
mheap: Do we need a response if it's correct? Would a `204` be enough?

definitions/verify.v2.yml
```diff
@@ -0,0 +1,427 @@
...
+            - cancelled
+        account_key:
+          type: string
+          description: The API key that made this Verify Request (may include subaccounts)
+          example: "acbd1234"
+        request_cost:
+          type: number
+          description: The one-off cost for this request, in EUR
+          example: "0.05"
+        messages_cost:
```
mheap: The messages API shows this inside a `usage` field

```
  "usage": {
    "currency": "EUR",
    "price": "0.0333"
  },
```

Do we want to split out base cost, messages cost, TTS cost etc? How about showing a `total_cost` field?

definitions/verify.v2.yml
```diff
@@ -0,0 +1,427 @@
...
+          type: number
+          description: The per-message/call costs for this request so far, in EUR. This number may update as the request progresses.
+        date_submitted:
```
mheap: Our date fields should use the `_at` suffix e.g. `submitted_at`

definitions/verify.v2.yml
```diff
@@ -0,0 +1,427 @@
...
+          type: string
+          description: Current status of this request
+          example: in_progress
+          enum:
+            - in_progress
+            - completed
+            - failed
+            - expired
+            - cancelled
+        account_key:
```
mheap: We call this `api_key` everywhere else

definitions/verify.v2.yml
```diff
@@ -0,0 +1,427 @@
...
+                type: string
+                description: The communication type
+                example: "sms"
+                enum:
+                  - sms
+                  - voice
+                  - whatsapp
+                  - viber_service_msg
+                  - messenger
+              timestamp:
```
mheap: What timestamp is this? `sent_at`, `received_at`?

definitions/verify.v2.yml
lornajane: It's a noun rather than a verb

definitions/verify.v2.yml
lornajane: It does, but one is the Voice API and one is the Messages API and they might do different things in future or something? Also the fields kind of have different descriptions although they're called the same thing. However I think we might be just-messages for the first phase so this may be something we don't need to decide now!

definitions/verify.v2.yml
lornajane: I don't think we know the total cost at this point in the process, because there can be a race condition between whether we are doing another operation at the same time as the pin is confirmed ... gonna leave this open though and figure out if we can improve on this

definitions/verify.v2.yml
lornajane: Although, other noun suggestions are also welcome ... status makes sense from a restful point of view but I wonder if we could be even more descriptive with what we pick here. Answers on a postcard!

definitions/verify.v2.yml
lornajane: I got it from Messages API but this is a great question! I kicked off discussion around this in the API standards channel,

definitions/verify.v2.yml
lornajane: A machine wouldn't need a response but I think some of our customers would be happier if there was one, if that makes sense. But I really agonised whether a 200 made sense for when there was nothing wrong with the request but the pin _isn't_ valid. I've got it here as an alternative response with a 200 status but maybe a failure status would make it more obvious that this isn't all fine?? Again, the 200 is technically correct but maybe misleading!

definitions/verify.v2.yml
```diff
@@ -0,0 +1,384 @@
...
+      properties:
+        type:
+          type: string
+          description: The type of message that you want to send.
+          example: 'sms'
+          enum:
+            - sms
+            - viber_service_msg
+            - whatsapp
+        number:
```
mheap: If we were to match this up to messages/dispatch, `number` would become `to` and `sender` would become `from`.

`number` is misleading as depending on the channel, it may be an ID. Even for things like whatsapp, although it looks like a phone number it's a string identifier

lornajane:
Closing this to await changes landing in Messages/Dispatch and then will regroup on it.

lornajane:
We're ready to look at this again, opening the PR to give the rendered preview

definitions/verify.v2.yml
mheap: I've just re-reviewed and hit this again. Verify is more RPC than REST, so I think using a verb here makes sense

definitions/verify.v2.yml
```diff
@@ -0,0 +1,384 @@
...
+          description: Request found and PIN checked
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/PinResponse'
+              examples:
+                success:
+                  summary: PIN accepted
+                  value:
+                    status: OK
```
mheap, line 84: I'm not sure what this response body adds. The `2xx` indicates that it's ok. Is there any more metadata we could add (e.g. last event triggered?) or should it be a `204`?

mheap: Actually, I've just seen the conversation above 🤦 

At least I'm being consistent!

definitions/verify.v2.yml
mheap: Coming back to this, the WhatApp deployment API is returning a `422` if your PIN is invalid. I think that would work here if we think of the PIN as validation. If it's successful, no further events are sent. If it's not, they get a validation error

definitions/verify.v2.yml
mheap: Based on a discussion a few weeks ago this might be `account_id`, but I'll confirm as soon as there's consensus

definitions/common/common_data.yml
```diff
@@ -0,0 +1,46 @@
+---
+components:
+  schemas:
+    Language:
+      type: string
+      description: Languages that are available to use
+      example: es-es
+      default: en-us
+      enum:
+        - ar-xa
```
Dbasdeo, line 10: The list of langs to support will be much smaller to start with:
English (en-us)
English (en-gb)
Spanish - Spain (es-es)
Spanish - Mexico (es-mx)
Spanish - US (es-us)
Italian (it-it)
French - France (fr-fr)
German (de-de)
Russian (ru-ru)
Hindi (hi-in)
Portuguese - Brazil (pt-br)
Portuguese - Portugal (pt-pt)
Indonesian (id-id)

definitions/verify.v2.yml
```diff
@@ -0,0 +1,612 @@
...
+            application/json:
+              schema:
+                required:
+                  - request_id
+                type: object
+                properties:
+                  request_id:
+                    $ref: '#/components/schemas/RequestId'
+        "422":
+          description: The request could not be found
```
Dbasdeo, line 37: This seems wrong

definitions/verify.v2.yml
```diff
@@ -0,0 +1,612 @@
...
+                    properties:
+                      title:
+                        description: Generic error message
+                        example: There was a problem with your submission
+                      type:
+                        description: Link to error / remediation options
+                        example: 'https://developer.nexmo.com/api-errors/verify#invalid-parameters'
+                      detail:
+                        description: Additional information about the error
+                        example: Missing required field `to`
```
Dbasdeo, line 54: There is no `to` field at the minute.

definitions/verify.v2.yml
```diff
@@ -0,0 +1,612 @@
...
+    post:
+      operationId: checkPIN
+      summary: Check a supplied PIN against a request to see if it is valid
+      description: Allows a PIN to be checked against an existing Verification request
+      requestBody:
+        $ref: '#/components/requestBodies/CheckPINRequest'
+
+      responses:
+        "204":
+          description: No content
```
Dbasdeo, line 118: I want to change this to 200 and return a small object like so:
```
{
 "number_of_events":"2",
 "price":"0.05"
}
```

definitions/verify.v2.yml
pardel: @Dbasdeo in what way?

Dbasdeo: 422 - The request could not be found

Sounds like a 404

definitions/verify.v2.yml
pardel: @Dbasdeo  As a rule, a delete operation should not return content. Why add pricing here?

Dbasdeo: This isn't a delete operation, its a post that checks the pin. As far I understand customers find these pieces of information extremely useful.

definitions/verify.v2.yml
```diff
@@ -0,0 +1,606 @@
...
+    #   type: integer
+    #   description: How long the generated PIN code is valid for
+    #   minimum: 60
+    #   maximum: 3600
+    #   default: 300
+    #   example: 300
+
+    PINLength:
+      type: integer
+      description: Length of the PIN to send to the user
```
Dbasdeo, line 368: The pin length should be 4-10

definitions/verify.v2.yml
pardel: @Dbasdeo you're absolutely right. Go for it!

definitions/verify.v2.yml
pardel: updated

definitions/verify.v2.yml
```diff
@@ -0,0 +1,611 @@
...
+                    example: abc:123:verify
+                required:
+                  - channel
+                  - to
+              example: [
+                  {
+                    "channel": "whatsapp",
+                    "to": "447700900000",
+                    "from": "447700900000",
+                    "template": "abc:123:verify"
```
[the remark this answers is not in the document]
pardel: Also, they shouldn't care what WhatsApp template we're using :)

definitions/verify.v2.yml
```diff
@@ -0,0 +1,611 @@
...
+      default: 4
+      example: 4
+      enum:
+        - 4
+        - 10
+
+    RequestId:
+      type: string
+      description: 'The ID of the request'
+      example: '0123456789abcdef0123456789abcdef'
```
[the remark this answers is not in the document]
pardel: @Dbasdeo Sure. Can you supply an example one?

definitions/verify.v2.yml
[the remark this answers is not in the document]
Dbasdeo: This was meant to be a future option when a customer uses their own number and then has to provide their own template.

definitions/verify.v2.yml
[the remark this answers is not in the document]
Dbasdeo: "c11236f4-00bf-4b89-84ba-88b25df97315"

definitions/verify.v2.yml
Dbasdeo: This still says 204? We have this implemented as 200

definitions/verify.v2.yml
Dbasdeo: ? This doesn't seem like it was fixed?

definitions/verify.v2.yml
```diff
@@ -0,0 +1,609 @@
...
+                type: object
+                properties:
+                  channel:
+                    type: string
+                    enum: 
+                      - whatsapp
+                    example: whatsapp
+                  to:
+                    $ref: '#/components/schemas/To'
+                  from:
```
grquintero, line 284: Our customers don't need to specify 'from' for the WhatsApp channel.

SecondeJK:
This is ready to go.

pardel (approved):
LGTM

PR #226, opened 2019-12-10
Fix typo in definitions/media.yml

timcraft:
Assuming this is a typo in the specification, not in the API itself.

Not sure whether version should be incremented? Can this be combined with any other changes?


tbedford:
@fabianrbz: not sure if the build failure is important here, so I have not merged.

PR #238, opened 2020-01-07
Added link to list of supported languages

pardel:
# Description

* Added link to the FB page containing list of supported languages

### DIFF
--- a/definitions/dispatch.yml
+++ b/definitions/dispatch.yml
@@ -270,7 +270,7 @@ components:
                 - fallback
                 - deterministic
             locale:
-              description: 'We are using the industry standard, BCP 47, for locales. So in your API call to the /messages API you will need to put “en-GB” and this will refer to the “en_GB” template for WhatsApp.'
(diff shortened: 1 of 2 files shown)


lornajane:
I know this was only a copy change but in general, I would prefer us to use the pull request template and increment the version numbers on the edited specs regardless. No need to go back and do anything with this one but in general - the pull request template is there for a reason on this repo.

tbedford:
I always remember to fill out the template on my own PRs and I always seem to forget when reviewing other peoples because when I come to do the review the template is already gone! Is there something we can do to stop that happening besides providing me with a new brain?

lornajane:
Not sure. I think the real problem is that we do have PR templates in other places and we ignore them there! I am also reluctant to automate this because there are lots of valid reasons for not needing exactly that template, for example if making a change to the wider repo and not the spec. I saw it in the email notification but didn't comment until I could see the diff and of course you beat me to the review :)

tbedford:
OK thanks I will try and remember to catch this in future.

PR #240, opened 2020-01-16
Add a spectral function to check format of parameter names

lornajane:
# Description

Use a spectral custom function to check parameter naming.

# New 

* Add parameter format checking for API standards

# Checklist

### DIFF
--- a/.spectral.yml
+++ b/.spectral.yml
@@ -1,4 +1,5 @@
 extends: spectral:oas
+functions: [nexmo-property-name-underscores]
 rules:
   example-value-or-externalValue: false
   oas3-unused-components-schema: false # needed by inter-file references e.g. Voice API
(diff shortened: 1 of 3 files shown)


lornajane:
This works for request parameters but I'm struggling to access response names because those are keys in the data structure, not values! Got a question on the support forum and other advice is gratefully received.

We have a few specs that don't meet these requirements but I've set it to warn so we'll never see it unless we validate directly with spectral. The main limitation is the rule about one underscore (from here https://nexmo.github.io/api-standards/standard/property-names/)

lornajane:
Also, sorry, I seem to have broken something but I am not sure how to tell what ...

fabianrbz:
@lornajane i've re-run the action and now it passed, I'm not sure what happened there.
is this ready for review?

lornajane:
It's nice that the build passes but I'm not sure what happened there! Yes I think this is ready to go

PR #249, opened 2020-02-04
Indicate that the IP Address field isn't used in Verify any more

lornajane:
# Description

The IP Address field is no longer used, make this clearer in the reference docs

# Fixes

* Clarify that the `ip_address` field isn't expected/used

# Checklist


lornajane:
**Warning** be aware that there are multiple PRs open for Verify so while the version is incremented here, it might already be incremented. Please check master before merging!

PR #251, opened 2020-02-06
Check for valid version strings

### LINKED ISSUE #235: Ensure that the `version` key is a valid SemVer version
This includes making sure that it is a string and has three segments. A two segment version is cast as a float in YAML and breaks some tooling

lornajane:
# Description

Fixes #235 by checking the versions follow a x.x.x format.

# New 

* Validate format of specs' version field

# Checklist

^ yes I updated the conversation.v2 spec and incremented the version

### DIFF
--- a/.spectral.yml
+++ b/.spectral.yml
@@ -26,3 +26,13 @@ rules:
     then:
       function: "nexmo-property-name-underscores"
 
+  # Require 3-part Semantic Versions as the spec versions
+  semver:
+    severity: error
+    recommended: true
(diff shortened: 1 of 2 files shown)


lornajane:
I've set this to severity error - but `conversation.v2.yml` then fails as it has `beta2` as its version string. Should we update this as well with this change? I'm not sure what version would make sense.

mheap:
Let’s update that to 0.2.0 - that means beta 2 based on our API standards

PR #269, opened 2020-02-24
Move Verify API over to use POST methods

lornajane:
# Fixes

* Switch Verify API endpoints to use POST requests

# Checklist

### DIFF
--- a/definitions/verify.yml
+++ b/definitions/verify.yml
@@ -31,9 +31,9 @@ security:
   - apiKey: []
 paths:
   "/{format}":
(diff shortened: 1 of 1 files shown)


lornajane:
(this one still a bit of a work in progress, I need to do a bunch more testing to make sure it all really does make sense)

lornajane:
I'm not sure who would be best to review, feel free to tag others but we've had a number of customer comments related to this so I'd love to get it moving if it seems solid.

PR #276, opened 2020-03-10
Add api-key field to inbound-sms request

tbedford:
# Fixes

* Adds `api-key` field to Inbound SMS request body. 

# Checklist

### DIFF
--- a/definitions/sms.yml
+++ b/definitions/sms.yml
@@ -1,6 +1,6 @@
 openapi: "3.0.0"
 info:
-  version: 1.0.5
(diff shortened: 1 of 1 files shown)


lornajane:
Can I just check: was there a reason this was opened in preference to the existing https://github.com/Nexmo/api-specification/pull/273 PR which looks like it did the same?

tbedford:
@lornajane yes, I looked at that PR and it was to do with DLRs, mine was a quick fix for Inbound SMS as that was the fix required for the ticket that had been raised.

PR #278, opened 2020-03-16
Pick up some data type improvements for Verify API

lornajane:
# Description

Update some strings in the example fields that I spotted when working with Prism

# Fixes

* Make the string fields quoted strings in examples

# Checklist


lornajane:
Abandoning this, I merged it into #269 so we might as well review it all at once

PR #279, opened 2020-03-17
Making search_pattern description a little clearer

slorello89:
Thinking that this description should perhaps change to note that the country code is inclusive for value `0` - for US developers, in particular, we tend to think in terms of our area codes (321, 201, 305, etc. . .) And as I stumbled on this one I think others might be confused by it as well.

# Description

# Checklist


slorello89:
Thanks @lornajane - good point - I just made a quick edit (US starts with '1' not '0') :)

PR #283, opened 2020-03-31
Add recent API changes to the Numbers spec

lornajane:
# Description

We made changes to the Numbers API, this updates the spec to match

# Fixes

* Add missing updates from recent API change

# Checklist


lornajane:
Good spot, I went with Application in more places and I think that does make more sense.

PR #284, opened 2020-03-31
Fix OAS_PATH for Create Review Apps.

fabianrbz:
# Description

Fix `OAS_PATH` for create review apps


benhayehudi:
I was too slow to review for @mheap 😂

PR #286, opened 2020-04-06
Add client-ref field, error info and better examples for SMS API

lornajane:
# New 

* Describe error response to Send an SMS

# Fixes

* Add missing (optional) client-ref field to API response and DLR payloads

# Checklist

### DIFF
--- a/definitions/sms.yml
+++ b/definitions/sms.yml
@@ -255,9 +255,17 @@ components:
         name: mt-submission-response
       properties:
         messages:
+          x-skip-response-description: true
           type: array
           items:
-            $ref: '#/components/schemas/Message'
+            $ref: "#/components/schemas/Message"
+          properties:
+            count:
+              type: integer
+              example: 1
+              xml:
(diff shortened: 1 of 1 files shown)


definitions/sms.yml
```diff
@@ -345,26 +371,14 @@ components:
...
-        name: mt-submission-response
-      properties:
-        messages:
-          x-skip-response-description: true
-          type: array
-          items:
-            $ref: '#/components/schemas/Message'
-          properties:
-            count:
-              type: integer
```
mheap, line 364: We seem to have lost this in the refactor, but it exists in the API response

lornajane: Ah, I remember this. I ran into an issue figuring out how to express this correctly while still including the error definition (because the attribute is at a level that's sort of in the way). If you can fix it, that would be grand but I hit a wall with it! Another option is to capture this as an issue, merge what we have, and iterate. Your call!

PR #290, opened 2020-04-17
To number is a string

tbedford:
# Fixes

* `to` number is a string but was displaying as number in response examples.

# Checklist

### DIFF
--- a/definitions/developer/messages.yml
+++ b/definitions/developer/messages.yml
@@ -1,7 +1,7 @@
 openapi: 3.0.0
 info:
   title: Message Search API
-  version: 1.0.3
+  version: 1.0.4
   description: >-
     The Messages API lets you retrieve messages you have sent via the SMS API by
     ID, as well as retrieve details of messages that were rejected.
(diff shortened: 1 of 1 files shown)


tbedford:
@slorello89 Thanks Stephen, sadly there are some weird build errors I'll need to look into!

mheap:
@tbedford Passes locally for me too. If we mark this as ready I can force merge

tbedford:
@mheap looks like there is an issue but not to do with my change.

lornajane:
Looks like some sort of spectral dependency error:  https://github.com/stoplightio/spectral/issues/1097 I'll keep looking but I agree it's not caused by this PR directly

lornajane:
Ah, it's fixed, this was just bad timing as they had a dependency version problem a couple of days ago. I can't see how to kick off the travis build again but apparently closing and re-opening the PR does it ....

lornajane:
So how about now?

lornajane:
Merged, thanks for picking up on the data type of the example @tbedford - it's the little things that make the biggest difference :)

PR #292, opened 2020-04-21
Add detail of responses for the error status codes to Numbers API

lornajane:
# New 

* Describe error responses in Numbers API

# Checklist

### DIFF
--- a/definitions/numbers.yml
+++ b/definitions/numbers.yml
@@ -1,7 +1,7 @@
 openapi: 3.0.0
 info:
   title: Numbers API
-  version: 1.0.18
+  version: 1.0.19
   description: >-
     The Numbers API enables you to manage your existing numbers and buy new virtual numbers for
     use with Nexmo's APIs. Further information is here: <https://developer.nexmo.com/numbers/overview>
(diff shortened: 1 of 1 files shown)


definitions/numbers.yml
```diff
@@ -58,16 +58,10 @@ paths:
...
-                    description: A code relating to this error
-                    example: "401"
-                  error-code-label:
-                    type: string
-                    description: Words describing the error that occurred
-                    example: authentication failed
+                $ref: "#/components/schemas/unauthorized"
+            text/xml:
+              schema:
+                $ref: "#/components/schemas/unauthorized"
```
mheap, line 64: These seem to be missing `xml.name` which controls the name of the wrapper. In the API it's `response` but in the rendered version it's `hash`

lornajane:
This was a good catch but it was a bit of a can of worms since most of the responses were missing the XML root tag info! Added for all endpoints, rebased the branch and bumped the version number. This should be in good shape now!

definitions/numbers.yml
lornajane: Good spot. This was already missing from the other endpoints too which I don't think helped but I've added them all now!

mheap:
🙌 Thanks for seeing this through

PR #298, opened 2020-05-11
[WIP] Use the except block to "permit" style violations

lornajane:
# Description

We have some style checking rules disabled or made minor because not all of our APIs meet the standards. Using the new `except:` block lets us add exceptions for existing violations and still detect anything new that is added.

Requires spectral 5.2.0 or later.

# New 

* Increase level of standards checks

# Checklist

^ no, no spec changes


lornajane:
not needed

PR #299, opened 2020-05-12
Add PSD2 endpoint to Verify API

lornajane:
# Description

We have added a feature to enable secure payments confirmation for online shopping (an SCA method for the PSD2 directive). Add this to the spec.

# New 

* Added new Verify API PSD2 feature for confirming payments

# Checklist

### DIFF
--- a/definitions/verify.yml
+++ b/definitions/verify.yml
@@ -1,9 +1,9 @@
-openapi: 3.0.0
+openapi: 3.0.3
 servers:
   - url: "https://api.nexmo.com/verify"
 info:
   title: Nexmo Verify API
-  version: 1.1.0
+  version: 1.1.1
   description: >-
     The Verify API helps you to implement 2FA (two-factor authentication) in your applications.
     This is useful for:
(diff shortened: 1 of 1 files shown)


definitions/verify.yml
```diff
@@ -645,6 +645,156 @@ paths:
...
+        content:
+          application/x-www-form-urlencoded:
+            schema:
+              type: object
+              required:
+                - api_key
+              properties:
+                api_key:
+                  $ref: "#/components/schemas/api_key"
+                api_secret:
```
grquintero, line 678: api_secret and number would be "required" as well, no? And I don't recall if "payee" and "amount" are required as well under the PSD2 endpoint.

lornajane: Ahhh! Good catch :)

lornajane:
I've updated to make sure the relevant fields are marked as required. If any of you are happy enough to approve this change, then I can merge and start the processes for SDKs and code snippets.

lornajane:
Chatted with @grquintero who is happy that this is good to go

PR #301, opened 2020-05-13
Athena oas

slorello89:
# Description
Draft PR for the NLU Open API Spec

# Checklist

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+    "cSpell.words": [
+        "Vonage's"
+    ]
+}
--- a/definitions/athena.yml
+++ b/definitions/athena.yml
@@ -2,36 +2,191 @@ openapi: 3.0.0
 info:
   version: 0.1.0
   title: Natural Language Understanding API
-  description: "Vonage Natural Language Understanding API allows you to build, store, and use model's for intent \nand entity detection from user utterances.\n"
+  description: "Vonage's Natural Language Understanding API allows you to build, store, and use model's for intent \nand entity detection from user utterances.\n"
   contact:
     name: Nexmo
     url: 'https://developer.nexmo.com/'
     email: devrel@nexmo.com
 servers:
   - url: 'https://api.nexmo.com/v0.1/athena'
-paths:
-  "/entities/{entity_iud}/values/{value_id}/synonyms":
+paths:  
+  
+  "/entities/{entity_id}/values/{value_id}/synonyms":
+    post:
+      operationId: create_synonym
+      tags:
+        - synonyms
+      summary: Create a synonym for a value of an entity
+      requestBody:
+        $ref: '#/components/requestBodies/Synonym'
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/synonym'
     get:
       operationId: get_synonyms
       tags:
-        - synonym
-      summary: List synonyms for an entity value
+        - synonyms
+      summary: List synonyms of an entity value
       parameters:
+        - $ref: "#/components/parameters/record_index"
         - $ref: "#/components/parameters/page_size"
         - $ref: "#/components/parameters/cursor"
         - $ref: "#/components/parameters/order"
       responses:
         "200":
+          description: OK
+          content:
+            application/json:
+              schema:
+                properties:
+                  page_size:
+                    type: integer
+                    example: 10
+                    description: The number of results returned on this page.
+                  record_index:
+                    $ref: '#/components/schemas/record_index'
+                  _embedded:
+                    type: object
+                    x-nexmo-developer-collection-description-shown: true
+                    properties:
+                      synonyms:
+                        type: array
+                        description: List of Synonyms matching provided filter
+                        items:
+                          $ref: "#/components/schemas/synonym"
+        "404":
+          description: Resource Not Found
+  "/entities/{entity_id}/values/{value_id}/synonyms/{synonym_id}":
+    parameters:
+        - $ref: '#/components/parameters/synonym_id'
+    get:
+      operationId: get_synonym
+      tags:
+        - synonyms
+      responses:
+        '200':
+          description: Synonym retrieved
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/synonym"
+        '404':
+          description: Synonym Not found
+    delete:
+      operationId: delete_synonym
+      tags:
+        - synonyms        
+      responses:
+        '204':
+          description: No Content
+        '404':
+          description: Synonym Not Found
+  "/entities/{entity_id}/values":
+    get:
+      operationId: get_values
+      tags:
+        - values
+      summary: List Values of an entity
+      parameters:
+        - $ref: "#/components/parameters/record_index"
+        - $ref: "#/components/parameters/page_size"
+        - $ref: "#/components/parameters/cursor"
+        - $ref: "#/components/parameters/order"
+      responses:
+        '200':
+          description: Values Listed
+          content:
+            application/json:
+              schema:
+                properties:
+                  record_index:
+                    $ref: '#/components/schemas/record_index'
+                  _embedded:
+                    type: object
+                    x-nexmo-developer-collection-description-shown: true
+                    properties:
+                      values:
+                        type: array
+                        description: List of Values matching provided filter
+                        items:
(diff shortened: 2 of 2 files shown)


mheap:
# General

* Which form of pagination are we choosing? Looks like cursor based. If so, `record_index` isn't a valid pagination key as you always have to start from the beginning/end
* What are the error conditions for each endpoint? This is especially important to make sure we're consistent with things like DELETE endpoints (which aren't idempotent. It should 404 after the first delete)
* Should we be able to delete intents/entities/models? What happens to GET /analyze/:id if a referenced entity is deleted?
* When referring to a primary key, there's no need for a prefix e.g. model_id => id in the path
* Some missing descriptions for parameters
* Endpoints should be ordered consistently (e.g entities is out of order). List, POST, GET, PATCH, DELETE
* Could we use _links on all entities to aid discovery? e.g. can I filter models by providing an entity ID and see which models contain it? If so, I'd add that to _links on the model


# /analyze

* Are `model_id` and `intent_ids` exclusive?
* A model contains intents and entities. If I send `model_id` does it overwrite both?
* Why do we have a GET /analyze/:id endpoint if it's synchronous?


# /models

* Max page size can be expressed as an OAS constraint

# /intents

* The `oneOf` here threw me off. Need to fix that rendering bug
* Not sure on the naming of `start_idx` and `end_idx`
* POST /intents shouldn't have a `_links` section
* PATCH /intents/:id - Is this used to add entries to utterances, or to replace the entire array?

# /entities

* POST /entities - Do we have an example for `values`?
* `_links` section in POST and PATCH should be specific to that entity, not the generic pagination ones
* Top level on list endpoint is missing `_links`

# /synonyms

* Why is there no way to update/delete a synonym?
* Does the path need to be nested under /entities? A synonym is associated with a value, which appears to be globally unique. Can we express this as a top level resource of it's own, plus filters? See slides 38-47 https://speakerdeck.com/mheap/api-standards-2-dot-0-d518f960-c148-4a6f-be2d-6315ba1c9849?slide=38

mheap:
# General

* We're missing error definitions for all endpoints
* Some example still use default values e.g. abc123
* Are there opportunities to add more _links e.g. add a _links.utterances to an intent response?
* PATCH requests return 202. Should this be 200?
* When we say that "If set to true a delete will cascade to the resources children." can we explain which children they are?
* Endpoints should be listed in a consistent order: List, Create, Get, Patch, Put, Delete
* Single resource _links should only have self link
* If possible, have id be the first field returned in a response example


# Intents

* PATCH intent allows you to provide utterances but POST doesn't. Is this intended?

# Utterances

* I'm not sure what an utterance is from the description or example request/response
* id and _links should not be provided in the request body
* start_idx and end_idx still feel wrong to me. I'm inferring that idx is index, but is it 0 or 1 indexed and does it refer to bytes, characters or words?
* Why are we providing both entity_id and entity_name in the request? (Same for value and synonym)
* In fact, the whole `labeled_entities` section confuses me. I'm not sure what I'm looking at

# Values

* Do we want values to live inside entities? We discussed having entity_id be a required body parameter. This is only possible if value IDs are globally unique. (Reading further this looks like a mistake as PATCH has value at the top level)
* Synonyms are accepted in the request body, but not returned in the response
* You can set synonyms on POST but not on PATCH. If this is intended, let's add a note to the description

# Synonyms

* Should there be a bulk-add synonym endpoint? Should this be an update to a Values endpoint instead?

slorello89:
Pushing for a second round of comments:

`labeled_entities` are the entities that are part of the utterance provided - they're intended to help the model understand what entities can go along with a given intent.

`value` should be a component of entity as a value is meant to be a real-world manifestation of an entity

I don't think we should have a bulk upload for synonyms

mheap:
# General

* Check that all the examples make sense. e.g. Analyse endpoint talks about buying a red dress in the request, a tangerine in the response text and green mints in the intents + entities
* Explain which children will be deleted if you set the `force` parameter

# Analyse

* If model_id or intent_ids are mutually exclusive, that's a oneOf. Analyse with model, or analyse with intents

# Models

* _links uses ?int_id rather than _intent_id
* Is there a way to list all intents that are part of a model? If so, add to _links e.g. /intents?model_id=MOD-1234

# Utterances

* What is `text` used for in labeled_entities? Is this intended to match the value of either the value or synonym?
* Should there also be labeled_intents?

lornajane:
I'll start with the easy stuff that Spectral told me! Delete the rule exceptions (they are for older specs that I'm working through updating, we can get this one up to scratch right from the start) from `.spectral.yml` and run it. I see:
 * we should use `description` instead of `summary`
 * there are some entries in schemas that are unused
 * some of the examples don't have the correct data type.

Path entries: I think all the URL keys for paths should be strings.

Very good use of tags, these are really important in a longer doc. Check the tags section for consistent use of quotes (or not), consistent use of capitalisation on Utterances/Entities/Intents and also general sense (I spotted "Used to Analyze text for Intents an entities" which I think should be `s/an/and/` and probably with more or fewer capital letters)

With the error responses: these follow a standard structure. I think we can define that once and then use appropriate examples with each occurrence to make it more descriptive each time. (This one is a nice to have, it's perfectly serviceable and I don't want to block anything so feel free to capture into an issue and loop back on this later)

`POST` to `/value/{value_id}` is missing the `value_id` parameter. You can put the param after the path and before the verbs if it's the same for them all (spectral didn't get this, I use another validator that did).

_(this is already an essay, just starting to look at the rendered specs, will move to a new comment)_

lornajane:
For "Analyze Text"
 - if you have to supply _either_  `text` and `model_id` _or_ `text` and `intent_ids` then let's use `OneOf` to denote that
 - where do I get the IDs to supply for models, intents and entities?
 - what is the default behaviour for optional language field?

Looking at "List Models" but this is in lots of places
 - the example should be only the value of the field, not the key. So for example page size is an integer. The default is `10` and for the example we could choose ... `25` to show it can be changed.

For "Create an Utterance":
 - the fields documented for `labeled_entities` don't match what's shown in the request sample for this request

For "Delete an Entity" (and in at least one other place):
 - the description for 409 seems long and out of place here. Should it be just "Conflict" with the detail in the error repsonse itself?

I think I'd put the cursor field last everywhere, since users aren't expected to use it.

It would be brilliant if we could get a tech writer on this as well to go through looking at more descriptive wording, better formatting on sentence case and proper nouns etc. I absolutely love fruits and colours as example data but I'm confused about which one "tangerine" is so we could possibly rethink this!

lornajane:
One thing I'm left wondering is why we're using IDs instead of hypermedia everywhere. I'd expect all these things to be links, not IDs. That's not really related to the spec though!

slorello89:
Comments from Max and Valdemar:

Embed values in entities for both POST and PATCH
Embed Utterances in Intent POST/PATCH

axe synonyms endpoints - they don't really add much value - embed synonyms as list of strings in values requests

mheap:
# General

* In labeled_entities, do we need them to provide `text`? We already have `start_word` and `end_word` and can infer it
* Metadata of RGB: 00FF00 feels misleading. That's a hex colour, not RGB

# Analyze

* _embedded is at the wrong level. It should be _embedded.intents and _embedded.entities
* If possible, have the following order in responses: Endpoint specific fields, then _embedded, then _links

# Model

* A model should at least have an intent id or an entity id. <- Should this be a oneOf? I recall us discussing and saying that you can can create an empty model, but this feels wrong

# Intents

* Missing summary for GET /intent/:id
* Missing summary for DELETE /intent/:id
* 

# Entity

* Missing examples for values.name and values.metadata

# Values

* Missing example for metadata

slorello89:
@mheap - not quite sure I saw the missing examples for metadata you referred to. I restructured the embedded stuff - but insofar as putting links at the end - not quite sure what else there is to be done as links are part of each embedded object (maybe that's incorrect?). I agree with the model point - I added a switch

mheap:
> not quite sure I saw the missing examples for metadata you referred to

This was a bug in the renderer. All fixed 🙌 

>  but insofar as putting links at the end - not quite sure what else there is to be done as links are part of each embedded object

I _thought_ I saw `_links` being rendered in example responses before the actual data fields. Looking through now that's not the case, so feel free to ignore this

## General

* `INT-92a44c64-7e4e-485f-a0c4-1f2adfc44625` and `ENT-92a44c64-7e4e-485f-a0c4-1f2adfc44625` have the same suffix and could be confused
* In labeled_entities, do we need them to provide text? We already have start_word and end_word and can infer it

PR #302, opened 2020-05-13
adding note for WhatsApp Template charecter restrictions

slorello89:
# Description

OAS spec for Messages needs to be updated to reflect the new character restrictions on templates - these restrictions have been lifted from 160 chars total - to 60 chars in header / footer and 1024 chars in the body: https://developers.facebook.com/docs/whatsapp/business-management-api/message-templates/

# Checklist

### DIFF
--- a/definitions/messages-olympus.yml
+++ b/definitions/messages-olympus.yml
@@ -1,6 +1,6 @@
 openapi: "3.0.0"
 info:
-  version: 0.3.7
+  version: 0.3.8
   title: Messages API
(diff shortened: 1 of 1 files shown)


definitions/messages-olympus.yml
```diff
@@ -573,10 +573,13 @@ components:
           example: 'Additional text to accompany the image.'
     TemplateProperty:
       type: object
+      x-nexmo-developer-collection-description-shown: true
+      description: |
+        > **Note**: WhatsApp templates are restricted to having 60 characters in the header and footer components, and 1024 characters in the body component
```
tbedford, line 578: Should we say "60 characters **each** in the header and footer components"? Or is it 60 chars total?

slorello89:
The WhatsApp cluster is yet to be updated to handle the new restrictions so I'm tearing this PR down until we're ready.

PR #308, opened 2020-05-19
adding whatsapp provisioning API spec

slorello89:
# Description

Adding new WhatsApp provisioning API spec. This API will create a new WhatsApp cluster for the requester and perform WhatsApp OTP verification. Additionally, this API allows you to update your profile information/picture.

# New 

New WhatsApp provisioning API spec
# Checklist


mheap:
# General

* Missing example success responses
* Error conditions need adding
* How is the Profile initially populated?
* We should use example.com in all examples. business-email.com and img.com could resolve to something we're not expecting

# Deployment

* Some examples seem to have additional quotes as part of the example
* Should we accept `number` with and without the `country_code`?
* Rename `method` to `otp_method`?
* PATCH 202 should return the entity in the response
* current_state - should we use field names from our usual HTTP errors? state => title, description => detail

# OTP

* Should we allow the user to specify an OTP method when retriggering an OTP rather than using the one in the deployment?
* If so, rename `method` to `otp_method`?
* Return the method used to send the OTP in the resend-otp response
* Rename `method` to `otp_method` in the POST /verify endpoint too?
* Verify OTP HTTP code is 200 but the response says Accepted. Is this a synchronous or async method?

# Profile

* Validation rules for each field
* Is `vertical` case sensitive?
* Rename from `profile_photo_url` to `url` (the context is in the address)
* Should the endpoint be /profile not /profiles? You can only have one

# Profile Photo

* Should the endpoint be /photo not /photos? You can only have one
* Add restrictions from the description in to the description of the parameter
* Why does this return 201 when profile returns 200?

slorello89:
Addressed all of @mheap's comments anything that didn't have a direct action is reflected below:


General
Missing example success responses - which are still missing?
How is the Profile initially populated? - Once the OTP concludes a skeleton profile is populated

## Deployment

Should we accept number with and without the country_code? - Conclusion = no

## OTP

Should we allow the user to specify an OTP method when retriggering an OTP rather than using the one in the deployment? - It looks like we do already?

## Profile
Is vertical case sensitive? - looks like it is according to facebook's docs

mheap:
> Missing example success responses - which are still missing?

Update profile `200`

> Should we allow the user to specify an OTP method when retriggering an OTP rather than using the one in the deployment? - It looks like we do already?

The question was if this option should be removed. I don't think so, but wanted to understand the rationale for letting people choose the method again

> Is vertical case sensitive? - looks like it is according to facebook's docs

If might be for FB, but should it be for us? I'd like to see us be liberal in what we accept and then massage it in to the correct format for FB. We can do a straight passthrough if they send something we don't recognise


# Fix Deployment

* If you send a new `vname_certificate` will we respond with an `ERROR` current state, or would that change to pending?

# 404

```
  "detail": "Requested resource wasn't found"
```

This might not be technically true. It might exist, but they don't have access to it

# 422

Should contain `invalid_parameters` block as consumer clients expect it for validation errors

```
"invalid_parameters": [
    {
      "name": "code",
      "reason": "invalid value provided"
    }
  ],
```

```
"invalid_parameters": [
    {
      "name": "address",
      "reason": "is required"
    }
  ],
```

slorello89:
added response for update profile (also added a response to update profile picture)

I think allowing folks to request a different otp method is valid - what happens if an SMS is failing and you want to receive a phone call?

I went a different direction on vertical naming - let me know what you think. I axed the facebook names and replaced them with names that would be easily enumerable e.g. `Food And Grocery` became `food_and_grocery` I think don't think it's much of a burden on developers to ask for a particular set of string values - I do think it's worth making them more machine-readable - let me know what you think

Fix Deployment has not been agreed to yet - thus it shouldn't have been in there - thus I've removed it

I think the 404 should remain as is - If someone doesn't have access to a deployment - do we want to let them know it exists? if we do then we should add a 403

made updates for the 422

mheap:
> I think don't think it's much of a burden on developers to ask for a particular set of string values

Is `food_and_grocery` equivalent to `FOOD_AND_GROCERY` or `Food_and_Grocery`? I think they should be as we know their intent.

Switching to this format also removes our ability to do a straight passthrough if they send something that we don't understand. Do we want to ship a new version of the API each time they add a new vertical?

> I think the 404 should remain as is - If someone doesn't have access to a deployment - do we want to let them know it exists? if we do then we should add a 403

We don't change the message, we add to it. Here's an example from subaccounts:

```
  "detail": "API key 'acc6111f' does not exist, or you do not have access",
```

By indicating that it might exist (we don't confirm) but they don't have access it gives them slightly more information to work with.

---

## General

* Better `401` detail entry?
* `401` has `error` key rather than `title`
* Do we have an `api_key` value that we use in the docs consistently? If not, should we have one?

## POST /deployment

* I'm not sure that we should be showing `current_state: ERROR` as our default example

## POST /otp

* What value does the `message` field add on `202`?

## PATCH /profile

* I'd avoid using real places in examples. `123 Main Street` is a safe option

## PATCH /profile/photo

* Missing `invalid_parameters` on 422

mheap:
# General

> "detail": "Deployment does not exist or you do not have access.",

This error points to a generic error. I think it should point to a WA deployment specific error so that we can have specific resolution steps

* Instead of using the description, e.g `max 256 characters` use a maximum constraint on the schema

# POST /deployment

This example state doesn't look right to me:  
image

# PUT /profile/photo

We send `url` but respond with `source_url`

slorello89:
@mheap - RE max length: I can remove them from the description - but the three items that have max sizes are all constrained in the schema - it just doesn't get reflected in the renderer. Is this a renderer bug?

definitions/whatsapp-provisioning.yml
```diff
@@ -0,0 +1,736 @@
...
+        description: URL pointing to a .png or .jpg. Image must be minimum 192px. max 640px x 640px. max 5MB.
+        required: true
+        content:
+          application/json:
+            schema:
+              type: object
+              required: 
+                - source_url
+              properties:
+                source_url:
```
mheap, line 278: What clarity does `source_` add here? Does `source_url` also make sense in a `GET`?

definitions/whatsapp-provisioning.yml
```diff
@@ -0,0 +1,736 @@
...
+      required:
+        - type
+        - title
+        - detail
+        - instance
+      properties:
+        type:
+          type: string
+          description: Link to error / remediation options
+          example: https://developer.nexmo.com/api-errors#not-found
```
mheap, line 485: This should point to /api-errors/whatsapp-provisioning#not-found

Check all the error URLs and make sure that they're scoped correctly

definitions/whatsapp-provisioning.yml
slorello89: I changed the name to image_url - which is what this is.

definitions/whatsapp-provisioning.yml
slorello89: Done

definitions/whatsapp-provisioning.yml
```diff
@@ -0,0 +1,795 @@
...
+      required:
+        - type
+        - title
+        - detail
+        - instance
+      properties:
+        type:
+          type: string
+          description: Machine readable error type
+          example: https://developer.nexmo.com/api-errors/whatsapp-provisioning#unauthorized
```
mheap, line 568: Are auth errors specific to WA provisioning, or should we use the default instance of this?

definitions/whatsapp-provisioning.yml
```diff
@@ -0,0 +1,795 @@
...
+      required:
+        - type
+        - title
+        - detail
+        - instance
+      properties:
+        type:
+          type: string
+          description: Link to error / remediation options
+          example: https://developer.nexmo.com/api-errors/whatsapp-provisioning#VerificationFailed
```
mheap, line 431: kebab-case for fragments please (this is missing from the standards) e.g. `verification-failed`

definitions/whatsapp-provisioning.yml
```diff
@@ -0,0 +1,795 @@
...
+      required:
+        - type
+        - title
+        - detail
+        - instance
+      properties:
+        type:
+          type: string
+          description: Link to error / remediation options
+          example: https://developer.nexmo.com/api-errors/whatsapp-provisioning#unprocessable
```
mheap, line 506: This should be a separate error so that we can point them at the correct OAS endpoint that contains a valid request

definitions/whatsapp-provisioning.yml
```diff
@@ -0,0 +1,795 @@
...
```
mheap, line 469: What was unprocessable here?

mheap (dismissed):

laurenannmarie:
Hey, please could we change "method_otp" to "method" - this is the current implementation we have and in the interest of time, given WA's future roadmap or making this part of their new embedded sign up im not sure it's worth changing at this point?

slorello89:
> Hey, please could we change "method_otp" to "method" - this is the current implementation we have and in the interest of time, given WA's future roadmap or making this part of their new embedded sign up im not sure it's worth changing at this point?

@mheap I think it's ok to change back - thoughts?

laurenannmarie:
Hey! I think we are missing response code 201 from the POST - which means - no action is needed (2FA won't be triggered). Could this be added please?

definitions/whatsapp-provisioning.yml
```diff
@@ -0,0 +1,795 @@
...
+          description: Machine readable error type
+          example: https://developer.nexmo.com/api-errors/whatsapp-provisioning#unauthorized
+        title:
+          type: string
+          description: Error title
+          example: You did not provide correct credentials.
+        detail:
+          type: string
+          description: Unauthorized
+          example: Check that you're using the correct credentials, and that your account has this feature enabled
```
mheap, line 576: If the feature is not enabled it would be a `403` not a `401`

401 is purely for invalid authentication. If authorization fails after that point it's a 403

mheap:
I'd prefer `method_otp` but not enough to block a release

> Hey! I think we are missing response code 201 from the POST - which means - no action is needed (2FA won't be triggered). Could this be added please?

After chatting to @laurenannmarie it seems like this is actually an error case - when a deployment has already been triggered and is in progress. Waiting on Lauren to add more context

definitions/whatsapp-provisioning.yml
slorello89: I thought we decided that we specifically wouldn't send a 403 because we didn't want to indicate that it was a valid account?

definitions/whatsapp-provisioning.yml
slorello89: Added a more valid example

slorello89: Added a better example

definitions/whatsapp-provisioning.yml
slorello89: fixed for all of the errors

definitions/whatsapp-provisioning.yml
slorello89: added a more valid example

definitions/whatsapp-provisioning.yml
slorello89: Ok, split it out

slorello89:
> I'd prefer `method_otp` but not enough to block a release
> 
> > Hey! I think we are missing response code 201 from the POST - which means - no action is needed (2FA won't be triggered). Could this be added please?
> 
> After chatting to @laurenannmarie it seems like this is actually an error case - when a deployment has already been triggered and is in progress. Waiting on Lauren to add more context



> I'd prefer `method_otp` but not enough to block a release
> 
> > Hey! I think we are missing response code 201 from the POST - which means - no action is needed (2FA won't be triggered). Could this be added please?
> 
> After chatting to @laurenannmarie it seems like this is actually an error case - when a deployment has already been triggered and is in progress. Waiting on Lauren to add more context

@laurenannmarie I agree with @mheap that sounds like a 410 not a 201.

@mheap, I've reverted to `method` - I agree `otp_method` is better but I agree it should not be a blocker.

definitions/whatsapp-provisioning.yml
```diff
@@ -0,0 +1,800 @@
...
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorInvalidJson"
+        '401':
+          description: Invalid Credentials        
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorUnauthorized"
+        '404':
```
slorello89, line 134: This needs a 409 to be compliant with https://nexmo.github.io/api-standards/http-code/409/

definitions/whatsapp-provisioning.yml
```diff
@@ -417,6 +423,30 @@ components:
...
+          description: Link to error / remediation options
+          example: https://developer.nexmo.com/api-errors/whatsapp-provisioning#conflict-resend-otp
+        title:
+          type: string
+          description: Generic error message
+          example: Deployment Completed, Resend Failed
+        detail:
+          type: string
+          description: Additional information about the error
+          example: OTP could not be resent because the deployment has already been finalized.
```
mheap, line 445: This isn't necessarily true. We should return 409 if the deployment hasn't reached the `SEND_TOP` state yet too. This could occur before or after that state and be returned any time the resource is not in a valid state to resend an OTP

slorello89:
Need to test the WAM API in prod before pushing to master.

definitions/whatsapp-provisioning.yml
```diff
@@ -0,0 +1,838 @@
...
+        waba_id:
+          type: string
+          description: The WABA ID of the WhatsApp number
+        country_code:
+          type: string
+          description: the country code of the deployment e.g. `44`
+          example: '44'
+        number:
+          type: string
+          description: the Number for the deployment e.g. `7877001122`
```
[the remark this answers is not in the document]
marklewin: We should also mention omitting leading zeros?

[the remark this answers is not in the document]
slorello89: I think that's resolved by simply stating you leave out the international dial code, doesn't it?

marklewin (approved):
Thanks, LGTM

definitions/whatsapp-provisioning.yml
```diff
@@ -0,0 +1,838 @@
...
+          example: abcd1234
+        current_state:
+          type: object
+          description: The current status of the deployment
+          properties:
+            title:
+              type: string
+              description: The current status of the deployment.
+              example: INITIALIZING
+              enum:
```
avbelyaev, line 740: There are actually more WAM states - https://github.com/nexmoinc/whatsapp-manager/blob/master/src/main/java/com/nexmo/whatsapp/manager/number/config/DeploymentState.java 

@Dbasdeo do we hide states like MIGRATING/SCALING/etc somehow? Because otherwise, customers can be not prepared to face these states

PR #312, opened 2020-05-26
Add info for spectral to README, trim/update links

lornajane:
# Description

Update README to give some more current links and also add information about linting and style checking.

# New 

* Add documentation on the linting tool

# Checklist

^ No, this is only a readme change

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,19 +22,27 @@ These definitions provide a single point of truth that can be used end-to-end:
 
 ## Resources
 
-- [What is OpenAPI?](https://swagger.io/docs/specification/about/)
-- [A Visual Guide to What's New in Swagger 3.0](https://blog.readme.io/an-example-filled-guide-to-swagger-3-2/)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,19 +22,27 @@ These definitions provide a single point of truth that can be used end-to-end:
...
+
+### Check your API spec for validity and style
+
+We use [Spectral](https://stoplight.io/open-source/spectral/) to style and validity check our files. If you're using Stoplight Studio, this picks up our style definition by default. Otherwise, you can install the tool locally like this:
+
+`npm install -g @stoplight/spectral`
+
+And check your file (for example `verify.yml`):
+
+`spectral lint definitions/verify.yml`
```
slorello89, line 42: I know we have the action to run this automatically - but should we maybe add this to the PR checklist?

lornajane: I don't think so. More PR template means fewer pull requests in my experience and I wouldn't want to add any hurdles. Like you say, it'll happen anyway and a reviewer will look at it too so I think it's covered.

PR #319, opened 2020-05-29
Break Voice API objects out to components section

lornajane:
# Description

Working with the code generators has been great for making very speedy integrations with our APIs but I'd only done a few and I think I picked the easy ones! For Voice API, we didn't use the `components` section as much and it gave the generator a bit less to go on, creating files (and matching struct names) like:

```
model_inline_object.go
model_inline_response_200_1.go
model_inline_response_200_2.go
model_inline_response_200_3.go
model_inline_response_200_4.go
model_inline_response_200_5.go
model_inline_response_200_6.go
model_inline_response_200__embedded_calls.go
model_inline_response_200__embedded.go
model_inline_response_200__embedded__links.go
model_inline_response_200__embedded__links_self.go
model_inline_response_200.go
model_inline_response_200__links.go
model_inline_response_200__links_self.go
```

This refactor (moving objects to named schemas in the components section, also fixing one thing where we were missing `type: object` with a few objects) makes the `paths:` section more readable and also gives more for the tools to work with so my models are now called:

```
voice/model_create_call_response.go
voice/model_dtmf_response.go
voice/model_get_call_response.go
voice/model_get_calls_response.go
voice/model_start_stream_response.go
voice/model_start_talk_response.go
voice/model_stop_stream_response.go
voice/model_stop_talk_response.go
```

I don't think this should have side effects, does anyone have any worries about using this sort of structure? I think with this approved, I'd like to spend some time looking at whether we could improve the reuse on this without losing any of the specifics in it. Thoughts?

# Fixes

* Refactor structure of Voice API

# Checklist

### DIFF
--- a/definitions/voice.v2.yml
+++ b/definitions/voice.v2.yml
@@ -1,7 +1,7 @@
 ---
 openapi: 3.0.0
 info:
-  version: 2.1.0
+  version: 2.1.1
   title: Voice API BETA
   description: |
       This is the *beta* version of the Voice API. Calls created with v2 must be managed
--- a/definitions/voice.v2.yml
+++ b/definitions/voice.v2.yml
@@ -33,38 +33,7 @@ paths:
         content:
           application/json:
             schema:
-              required:
-                - to
-                - from
-                - event_url
-              properties:
-                to:
-                  type: array
-                  items:
-                    oneOf:
-                    - "$ref": "./voice.yml#/components/schemas/EndpointPhone"
-                    - "$ref": "./voice.yml#/components/schemas/EndpointApp"
-                    - "$ref": "./voice.yml#/components/schemas/EndpointSip"
-                    - "$ref": "./voice.yml#/components/schemas/EndpointWebsocket"
-                    - "$ref": "./voice.yml#/components/schemas/EndpointVBCExtension"
-                from:
-                  "$ref": "voice.yml#/components/schemas/EndpointPhone"
-                ncco:
-                  "$ref": "voice.yml#/paths/~1/post/requestBody/content/application~1json/schema/properties/ncco"
-                answer_url:
-                  "$ref": "voice.yml#/paths/~1/post/requestBody/content/application~1json/schema/properties/answer_url"
-                answer_method:
-                  "$ref": "voice.yml#/paths/~1/post/requestBody/content/application~1json/schema/properties/answer_method"
-                event_url:
-                  "$ref": "voice.yml#/paths/~1/post/requestBody/content/application~1json/schema/properties/event_url"
-                event_method:
-                  "$ref": "voice.yml#/paths/~1/post/requestBody/content/application~1json/schema/properties/event_method"
-                machine_detection:
-                  "$ref": "voice.yml#/paths/~1/post/requestBody/content/application~1json/schema/properties/machine_detection"
-                length_timer:
-                  "$ref": "voice.yml#/paths/~1/post/requestBody/content/application~1json/schema/properties/length_timer"
-                ringing_timer:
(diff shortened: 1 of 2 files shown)


lornajane:
Oops, I broke some of the between-files links. Let me take another look at that but feel free to comment on the overall approach

benhayehudi:
There is a small pending change in the create a call requirements waiting for review in https://github.com/Nexmo/api-specification/pull/317 that will impact the create a call component described here. Can we review and merge that one and then add the change here?

lornajane:
We definitely can! Do you want to comment on the overall approach here while we work on getting the ducks in the right order in the row?

benhayehudi:
> We definitely can! Do you want to comment on the overall approach here while we work on getting the ducks in the right order in the row?

I think #317 describes it 😄, but in short, `event_url` is only required if it is not set on an application level, so the PR clarifies that.

definitions/voice.v2.yml
```diff
@@ -33,38 +33,7 @@ paths:
...
-                  "$ref": "voice.yml#/paths/~1/post/requestBody/content/application~1json/schema/properties/event_url"
-                event_method:
-                  "$ref": "voice.yml#/paths/~1/post/requestBody/content/application~1json/schema/properties/event_method"
-                machine_detection:
-                  "$ref": "voice.yml#/paths/~1/post/requestBody/content/application~1json/schema/properties/machine_detection"
-                length_timer:
-                  "$ref": "voice.yml#/paths/~1/post/requestBody/content/application~1json/schema/properties/length_timer"
-                ringing_timer:
-                  "$ref": "voice.yml#/paths/~1/post/requestBody/content/application~1json/schema/properties/ringing_timer"
+              "$ref": "#/components/schemas/CreateCall"
```
mheap, line 36: If we're switching to refs, can we avoid using `components/schemas` for everything and use the named component types (e.g. `requestBodies` for this one)?

lornajane: I'm a bit confused about this, the only thing that can be included in `components/requestBodies` is a whole request body so I've done that here, is this what you had in mind (just the newest commit)? Are there other places you'd like this change?

mheap: Ah, I see the issue with using `requestBodies` now, We need to use `schemas` so that the tooling knows which object to create.

mheap:
Is this a good time to discuss component naming conventions?

`CreateCall` - is this a request or a response, or something else entirely?

lornajane:
There is no such thing as a bad time to discuss naming! I'd be really happy to both improve the naming on the voice specs and to think more widely about conventions that work well and we'd like to adopt in other specs as we work on them :)

I think here perhaps `CreateCallResponseBody` would have been better. What else can we improve while I'm here?

mheap:
I think this is a good starting point. We can use `requestBodies` to define the full response with content type and headers etc if needed. Perhaps `CreateCallRequestBody` and `CreateCallResponseBody`? I like the addition of `Body` as it makes it clear that it's not a full request object (with headers etc)

lornajane:
OK, I tried to make some of these things clearer, there could be more to do but if it's an improvement, let's merge. I'm holding off the Go SDK work until I know a bit more about the OpenAPI structure to expect.

lornajane:
(I think the build fail is because I updated two specs! Should I split this into two PRs?)

mheap:
Yep, the build is failing as two PRs were updated at the same time.

I think in this instance we can ignore it and do the magic steps afterwards to resolve any issues

lornajane:
Is there anything else needed here? Or can we merge and do "magic steps" (I'm not clear what that entails) to accept the changes.

mheap:
We can merge and do magic steps. I've left it open until I have time to do them straight after (it'll be this week). Proceed assuming that this is merged 👍

PR #326, opened 2020-06-19
Update description of From and To in getCall(s) for Voice API

lornajane:
# Description

The `GET /v1/calls` endpoint returns a different data structure (slightly!) than we describe in the spec. Spec says: array of objects with "from" and "number" properties. But I actually just get one of those objects, no array. This PR updates the spec to describe this.

# Fixes

* Improve the way we specify `from` and `to` in the "get calls" endpoint of Voice API

# Checklist

### DIFF
--- a/definitions/voice.yml
+++ b/definitions/voice.yml
@@ -1,7 +1,7 @@
 ---
 openapi: 3.0.0
 info:
-  version: 1.2.9
+  version: 1.2.10
   title: Voice API
   description:
     The Voice API lets you create outbound calls, control in-progress calls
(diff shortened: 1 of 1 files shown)


benhayehudi:
- Thank you for this! It's on my list for Monday!

slorello89:
Can we just validate that this has always been like this and this isn't something new? It looks to me like it probably always has been this way but I just want to make sure we aren't introducing something new.

lornajane:
Right, refactoring done, please take another look everyone. I'm conscious that we have other Voice API work in flight as well, please let me know if I can rebase or otherwise git wrangle in a helpful manner.

benhayehudi:
> Right, refactoring done, please take another look everyone. I'm conscious that we have other Voice API work in flight as well, please let me know if I can rebase or otherwise git wrangle in a helpful manner.

I see one merge conflict here, and not sure how to resolve it? It looks small, but small can be deceiving :)

lornajane:
ohhh, let me look what else was changed and rebase this so we know where we are ...

lornajane:
@benhayehudi Try this, it was the prettier changes because it reformatted the spec so git though there were changes in both branches. I've rebased it now so these changes are based after the prettier ones.

PR #334, opened 2020-06-24
Refactor Voice to use oneOf for use cases

mheap:
# Description

# Fixes

* Split `POST /calls` in to a oneOf for `answer_url` and `ncco` options
* Split `PUT /calls/:id` in to a oneOf for `transfer` with `answer_url` and `ncco` options, plus `hangup`, `mute`, `unmute`, `earmuff`, `unearmuff`
* Remove `dtmfAnswer` from the `from` option when placing a call
* Fix examples for `ncco`, `event_url` and `answer_url`

# Checklist


mheap:
@benhayehudi I think we could use a `minimum` and `maximum` constraint to express that in the spec

The review app bug is known with the current renderer. v2 (just released) doesn't have this issue

PR #337, opened 2020-06-29
Series of fixes captured in DEVX-3052

tbedford:
Fix issues as described in https://nexmoinc.atlassian.net/browse/DEVX-3052

# Fixes

* Various corrections to specification.

# Checklist

### DIFF
--- a/definitions/reports.yml
+++ b/definitions/reports.yml
@@ -1,6 +1,6 @@
 openapi: "3.0.0"
 info:
-  version: 2.1.5
+  version: 2.1.6
   title: Nexmo Reports API
   description: |
     [Nexmo's Reports API](/reports/overview) allows you to request a report of activity on your Nexmo account.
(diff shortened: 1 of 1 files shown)


definitions/reports.yml
```diff
@@ -104,31 +103,31 @@ paths:
           example: "2018-01-01T00:00:00+0000"
           in: query
         - name: include_message
-          description: Include the message contents in the records
+          description: Include the message contents in the records. Optional for use with products `SMS` and `MESSAGES`.
```
marklewin, line 106: To me, this makes it sound like you should include this field with all products, but that it is only optional for `SMS` and `MESSAGES`, which I don't think what point (8) in [the ticket](https://nexmoinc.atlassian.net/browse/DEVX-3052) is stressing. It is _only_ applicable for message products, but is optional.

tbedford: Good catch! Hopefully I addressed the other issue you mentioned too, and also made one other fix. Thanks!

PR #338, opened 2020-07-01
Removed wording that "Remaining Balance" was not being returned

### LINKED ISSUE #316: Weirdly worded description
https://github.com/Nexmo/api-specification/blob/6a9295721aeed4f89575400e35501c4fcaf1da0c/definitions/number-insight.yml#L218

The description here says that you won't get the `remaining_balance` back from the advanced async API - while the remaining balance is actually shown in the response object.

dragonmantank:
# Description
Fixes #316 

Changed the wording on NI regarding Remaining Balance. This might have just been a holdover from before. Ran various JSON and XML requests, and "remaining_balance" and "remainingBalance" were returned both in the Async payload as well as the confirmation message.

# Checklist

### DIFF
--- a/definitions/number-insight.yml
+++ b/definitions/number-insight.yml
@@ -215,7 +215,7 @@ components:
           example: "447700900000"
         remaining_balance:
           type: string
-          description: "Your account balance in EUR after this request. Not returned with Number Insight Advanced Async API."
+          description: "Your account balance in EUR after this request."
           example: "1.23456789"
           xml:
             name: remainingBalance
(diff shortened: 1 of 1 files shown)


marklewin:
@dragonmantank Have given a passing review and also bumped version number (which is required in the build step) but left for you to merge, in case @slorello89 has comments.

dragonmantank:
> and also bumped version number (which is required in the build step)

Thanks!

lornajane:
Why is this approved but not merged? Is there another dependency somewhere, and if so could we link to it?

dragonmantank:
Nope, just fell through the cracks. Merged.

PR #341, opened 2020-07-07
Add Subaccount error messages

marklewin:
# Description

* Added Subaccount error messages [DEVX-3457](https://nexmoinc.atlassian.net/browse/DEVX-3457)
* Based on list of errors provided [here](https://docs.google.com/document/d/1KFCAWZOpcKIxLTS3WVNqkRsFcyZ1HyVNg0MCe6wMLqs/edit?usp=sharing)

# Checklist

### DIFF
--- a/definitions/subaccounts.yml
+++ b/definitions/subaccounts.yml
@@ -665,18 +665,18 @@ components:
         - amount
 
 x-errors:
-  provisioning:
-    description: Maximum number of sub-accounts reached
-    resolution: Remove an existing subaccount before attempting to create a new one.
   validation:
     description: The request failed due to validation errors
     resolution: See the `detail` (and, if present, the `invalid_parameters`) fields for the exact reason this request was rejected and review the documentation.
     link:
       text: View API reference
       url: /api/subaccounts
-  internal-error:
-    description: An error has occurred in the platform while processing this request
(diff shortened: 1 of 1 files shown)


definitions/subaccounts.yml
```diff
@@ -663,3 +663,23 @@ components:
         - from
         - to
         - amount
+
+x-errors:
```
tbedford, line 667: I'll need some help with this one. From what I can make out this adds the following error messages:

1. Request failed due to validation errors
2. Invalid transfer
3. Maximum number of subaccounts reached
4. An error has occurred in the platform while processing the request.

However, when looking at the rendering, and clicking through the buttons for 400, 401,...., 422 I can only see the first in this list rendered, that is, 422 corresponds to 'Request failed due to validation errors'. I don't see where error messages 2, 3, 4  get rendered.

marklewin: There doesn’t appear to be any correlation between the errors reported here and the API reference - check out the SMS errors for example. I think it’s more a case of ‘if you get a response with the following error text in it, here is something sensible to do’.

mheap: For future reference, these errors show up on the `api-errors` page - see https://developer.nexmo.com/api-errors/subaccounts

tbedford: ah! Thanks @mheap

PR #300, opened 2020-05-12
Placeholder PR for updates to Messages API

lornajane:
# Description

# Checklist

- [ ] version number incremented (in the `info` section of the spec)

^ No, this description really needs editing before we're ready to merge!

PR #342, opened 2020-07-07
Tabbing NewMessage operation, making whatsapp required for MTM

slorello89:
# Description

Refreshing the Olympus OAS spec so that it can be a bit more usable then it is now. Making all channel/messageType permutations tabbable. Also made `whatsapp` node a required part of a WhatsApp message for MTM messages.

Example:

# New 

* Tab interface for `NewMessage` operation
* Made `whatsapp` node a required field for MTM
* Cleaned up descriptions so they only refer to the channel for the tab

# Fixes

* Minor fixes to descriptions.

# Checklist


mheap:
@slorello89 It looks like `Inbound Message` also needs updating

Also, if you provide the same `x-tab-id` for `Image`, `Template` etc it will automatically switch all subtabs when you select that value. So you can select FB->Image, and click WA and it'll have Image selected. Then if you click `Text` and go back to FB it'll have `Text` selected

dragonmantank:
Is there a way to re-deploy this for testing inside NDP/station?

lornajane:
I recreated the review app by removing the `review-app` label and then re-adding it (it helps to pause in the middle so GitHub gets both actions), does that help?

dragonmantank:
> I recreated the review app by removing the `review-app` label and then re-adding it (it helps to pause in the middle so GitHub gets both actions), does that help?

That does, thank you!

definitions/messages-olympus.yml
```diff
@@ -68,12 +68,12 @@ paths:
                 content:
                   application/json:
                     schema:
-                      $ref: "#/components/schemas/InboundMessage"
+                      $ref: "#/components/schemas/inbound-message"
```
dragonmantank, line 71: Why are we leaving this as hyphenated by camelCasing the others?

definitions/messages-olympus.yml
```diff
@@ -68,12 +68,12 @@ paths:
...
               responses:
                 "200":
                   description: Your server returns this code if it accepts the callback.
 
-components:
+components:
```
dragonmantank, line 76: extra whitespace

definitions/messages-olympus.yml
```diff
@@ -82,25 +82,1186 @@ components:
     basicAuth:
       type: http
       scheme: basic
-
   schemas:
-    NewMessage:
+    inbound-common:
```
dragonmantank, line 86: Same as above, but for all the names here. Most of the other APIs seem to use camelCase for schemas (though I noticed a few used PascalCase). For consistency we should at least stick to one for an API, and longer term clean up all of them to match.

lornajane:
I'm confused about the status of this PR. It has two approvals but neither one merged the changes. Also does it have any relation to the other updates in https://github.com/Nexmo/api-specification/pull/300?

For naming things, we kind of have gone mostly CamelCase for things inside a spec but there is no official style guide. +1 for keeping consistency within a spec for now I think.

slorello89:
> I'm confused about the status of this PR. It has two approvals but neither one merged the changes. Also does it have any relation to the other updates in #300?
> 
> For naming things, we kind of have gone mostly CamelCase for things inside a spec but there is no official style guide. +1 for keeping consistency within a spec for now I think.

Aye I agree - with consistency, at least within the spec - an oversight on my part.

dragonmantank:
> It has two approvals but neither one merged the changes.

From my end I almost never merge someone else's PR. I leave that up to the person submitting the PR or the one "most responsible" for finishing out the lifecycle of a code change.

definitions/messages-olympus.yml
```diff
@@ -1,6 +1,6 @@
 openapi: "3.0.0"
 info:
-  version: 0.3.8
+  version: 0.3.9
```
tbedford, line 3: Possible race hazard with version number specified in #359

tbedford (approved):
Apart from my separate comment on version number, looks good.

PR #349, opened 2020-08-03
Make visible each PUT request example from the VAPI spec

bencgreenberg:
# New

* Makes visible the schema for request example types already present in the spec

# Checklist

### DIFF
--- a/definitions/voice.yml
+++ b/definitions/voice.yml
@@ -1,7 +1,7 @@
 ---
 openapi: 3.0.0
 info:
-  version: 1.3.2
+  version: 1.3.3
   title: Voice API
   description:
     The Voice API lets you create outbound calls, control in-progress calls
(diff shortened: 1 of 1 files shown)


mheap:
```
  141:9    error  oas3-schema              Property `requests` is not expected to be here.
```

What problem is this PR trying to solve?

bencgreenberg:
> ```
>   141:9    error  oas3-schema              Property `requests` is not expected to be here.
> ```
> 
> What problem is this PR trying to solve?

Can you explain to me why adding `requests:` allows the various request examples to be displayed? @yallen011 asked if it was possible, and experimenting locally, I noticed adding that, enabled it. Why is that so? And what's the proper way to do so?

bencgreenberg:
@mheap has indicated the current behavior is actually a bug, and we're going to circle back on this tomorrow during working hours again. Noting it here for reference.

lornajane:
Redoc does show the examples so I think the spec is correct, and we may need to look at the tool to enable this for our own docs.

mheap:
The fix for the renderer is at https://github.com/Nexmo/nexmo-oas-renderer/pull/83

It was a race condition when trying to load within Station

PR #350, opened 2020-08-03
DEVX-3237 New TTS language and style parameters

vshisterov:
# Description

New talk operation parameters added: `language` and `style`

# Checklist

### DIFF
--- a/definitions/voice.yml
+++ b/definitions/voice.yml
@@ -966,7 +966,7 @@ components:
         - "vi-VN"
         - "yue-CN"
     style:
-      description: The voice style to use
+      description: The vocal style (vocal range, tessitura, and timbre) to use
       type: integer
       default: 0
       
--- a/definitions/voice.yml
+++ b/definitions/voice.yml
@@ -1,7 +1,7 @@
 ---
 openapi: 3.0.0
 info:
-  version: 1.3.2
+  version: 1.3.3
   title: Voice API
   description:
     The Voice API lets you create outbound calls, control in-progress calls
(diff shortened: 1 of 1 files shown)


definitions/voice.yml
```diff
@@ -912,107 +914,62 @@ components:
...
+        - "th-TH"
+        - "tr-TR"
+        - "uk-UA"
+        - "vi-VN"
+        - "yue-CN"
+    style:
+      description: The voice style to use
+      type: integer
+      default: 0
+
```
bencgreenberg, line 972: I think we also want to include an `enum` here of the possible values for `style`

vshisterov: It's not an enumeration, it's a plain integer.

bencgreenberg: what i meant, is we should include a list of all the possible value

vshisterov: It depends on the selected `language`. For integer parameters, OpenAPI doesn't suggest to have list of values, though we may add min and max, still the max. which is 11 currently, will be relevant for only one language (`en-US`).

bencgreenberg: Ah, okay, given its variability and inconsistency, then I see the point. Thanks, Victor.

bencgreenberg:
I'd like us to figure out what we are doing with `style` in the NDP PR before moving forward here.

definitions/voice.yml
```diff
@@ -912,107 +914,62 @@ components:
...
+        - "ru-RU"
+        - "sk-SK"
+        - "sv-SE"
+        - "th-TH"
+        - "tr-TR"
+        - "uk-UA"
+        - "vi-VN"
+        - "yue-CN"
+    style:
+      description: The voice style to use
```
bencgreenberg, line 969:
```suggestion
      description: The vocal style (vocal range, tessitura, and timbre) to use
```

PR #351, opened 2020-08-07
Update version of nexmo-oas-renderer

bencgreenberg:
# Description

## Misc

* Update version of nexmo-oas-renderer

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,3 @@
 source 'https://rubygems.org'
 
-gem 'nexmo-oas-renderer', '~> 2.0.0'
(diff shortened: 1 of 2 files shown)


definitions/voice.yml
```diff
@@ -797,7 +798,7 @@ components:
 
     EndpointApp:
       type: object
-      description: Connect to an App User
+      description: Connect to an App User (Voice API v2 Only)
```
mheap, line 801: This shouldn't be shown in v1. If we need to refactor v2 to use the child `oneOf` values let's do that.

If you'd like a hand, @lornajane or @dragonmantank are our go-to OAS specialists

PR #352, opened 2020-08-11
Update Endpoint Types for VAPI V2 Create Call

bencgreenberg:
# Description

## Fixes
* Adds `app` as possible endpoint types for a create call request in VAPI v2

# Checklist

### DIFF
--- a/definitions/voice.v2.yml
+++ b/definitions/voice.v2.yml
@@ -1,7 +1,7 @@
 ---
 openapi: 3.0.0
 info:
-  version: 2.2.0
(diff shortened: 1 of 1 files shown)


bencgreenberg:
> This looks solid! Approved! I think there is room for refactoring to reuse some of these fields between specs (and AppEndpoint is defined in a spec where it isn't used) but that's a nice-to-have and shouldn't block this change.

I'm going to add it to my list for a future PR to improve upon. Thank you, @lornajane!

PR #356, opened 2020-08-28
Add WebSocket Call product

vshisterov:
# Description

New report type and corresponding `product` value supported: `WEBSOCKET-CALL`.

# Checklist


bencgreenberg:
@tbedford, As the Reports API specialist, want to take a peek at this?

definitions/reports.yml
```diff
@@ -1322,6 +1359,13 @@ components:
       allOf:
         - $ref: "#/components/schemas/report_request"
         - $ref: "#/components/schemas/voice_call_fields"
+        
+    WEBSOCKET-CALL:
```
lornajane, line 1363: Why are these schema names uppercase? We don't have naming checks on these fields (or it only emits a warning, not sure), do you know if this is like this for a reason (I realise this one just follows the others)

lornajane:
@vshisterov I see this has been open for a while, what's the status of this feature?

definitions/reports.yml
vshisterov: I have no idea, to be honest - as you said, I was following what is already in the spec. "Bad Consistent > Good Inconsistent" :)

vshisterov:
> @vshisterov I see this has been open for a while, what's the status of this feature?

It's in production for more than a month already (the Reports API support for WS). The docs still just awaiting approval for the merge per my understanding.

tbedford (changes requested):
For 'get records' I don't see the product in the example response, however it's there for 'Create an asynchronous report'.

definitions/reports.yml
tbedford: It's probably because the product names are uppercase in the responses. I also just followed existing usage.

tbedford:
Sorry, for some reason this one fell through the cracks. I deal with a LOT of PRs and usually respond to them very quickly. Apologies this one got missed.

tbedford (approved):

PR #359, opened 2020-09-08
adding unsupported type to messages-olympus

slorello89:
# New 

* Adding unsupported type to inbound messages webhook for messages-Olympus. This will occur if a channel receives a type that's not supported (previous behavior was just to drop the message).

# Checklist

### DIFF
--- a/definitions/messages-olympus.yml
+++ b/definitions/messages-olympus.yml
@@ -1,6 +1,6 @@
 openapi: "3.0.0"
 info:
-  version: 0.3.8
+  version: 0.3.9
   title: Messages API
   description: "The Messaging API is a new API that consolidates all messaging channels. It encapsulates a user (developer) from having to use multiple APIs to interact with our various channels such as SMS, MMS, Viber, Facebook Messenger, etc. The API normalises information across all channels to abstracted to, from and content. This API is currently in Beta."
   contact:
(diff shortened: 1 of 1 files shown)


definitions/messages-olympus.yml
```diff
@@ -272,6 +272,7 @@ components:
                     **whatsapp** and **messenger** supports `text`, `image`, `audio`, `video`, `file` and `location`. WhatsApp maximum inbound size is 64mb.
                     **mms** supports `image`.
                     **viber** supports `text`.
+                    <br\>The type will be set to `unsupported` if the WhatsApp, Messenger, Viber, of MMS channel receive a message that is of a not supported type.
```
peterjurkovic, line 275: can we also add an example such a message?

slorello89: As in an example request in the right pane? Or just an inline description (e.g. a WhatsApp sticker)?

AFavreNexmo, line 275: Not sure if it is needed to list all channels.

AFavreNexmo (approved):

definitions/messages-olympus.yml
peterjurkovic: I meant code snipped of unsupported message. E.g.:

```json
{
  "message_uuid": "aaaaaaaa-bbbb-cccc-dddd-0123456789ab",
  "to": {
    "type": "messenger",
    "id": "0123456678901234"
  },
  "from": {
    "type": "messenger",
    "id": "0123456678901235"
  },
  "timestamp": "2019-09-12T13:43:16.611Z",
  "direction": "inbound",
  "message": {
    "content": {
      "type": "unsupported"
    }
  }
}
```

slorello89: That's something that could be added in #342. Don't think it makes sense to add in here (have to add a new tab for it which won't make sense since we don't have all the tabs for the currently supported types yet)

slorello89: Done

tbedford (approved):

kurilinavonage (approved):

PR #361, opened 2020-09-15
Number insight  refresh

lornajane:
# Description

While working with the Number Insights API I found a couple of really minor things so I fixed them.

# Fixes

* Adds the missing `error_text` field to the async response of Number Insight API
* Corrects the `remaining_balance` data type to be a string everywhere on Number Insight API

# Checklist


marklewin (approved):
LGTM

PR #362, opened 2020-09-23
removing accidentally added sms MO & swapping id/number for viber inbound

slorello89:
# Description

* Accidentally added SMS MO to previous spec refresh - removing as it's not supported by the Messages API yet.
* Also needed to swap the id/number for viber inbound messages.
# Checklist


marklewin (approved):
Looks good in review app

PR #363, opened 2020-09-24
Refreshing Applications API (and common errors)

lornajane:
# Description

Applications uses content from another file but when the API definition is hydrated with that referenced content, the resulting output isn't valid OpenAPI, so this fixes that - plus a few other minor things that I found when integrating with Go.

# Fixes

* Improved examples and data types, renamed the common errors component.

# Checklist

### DIFF
--- a/definitions/application.v2.yml
+++ b/definitions/application.v2.yml
@@ -38,31 +38,7 @@ paths:
           content:
             application/json:
               schema:
-                properties:
-                  page_size:
-                    type: integer
-                    example: 10
-                    description: The number of applications per page
-                  page:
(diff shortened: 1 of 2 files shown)


lornajane:
These are reasonably minor changes, but I'd still really appreciate a review. Any comments, @mheap @dragonmantank - or anyone else you think would be able to offer insight on this (delegation is always allowed!)?

mheap:
@lukeocodes and @dragonmantank are your two people here I think. Luke as applications specialist, Chris for OAI

lukeocodes (approved):
LGTM!

PR #364, opened 2020-09-28
Sms-MO for messages

slorello89:
# Description

Part of the SMS-mo design discussion - putting the sms-MO we've discussed into yml.

Example:

# New 

* Adds SMS-mo to messages API spec

# Checklist


lornajane:
I think the incoming SMS should be a webhook, not a callback, because the event it is a result of isn't the API call that it's listed as a callback of (this is true for all incoming messages, I'm not sure why this OpenAPI definition is structured this way). Since OpenAPI 3.1 isn't released yet, the best thing would be to use the not-quite-standard `x-webhooks` format we have in the SMS API and then upgrade the specs later.

dragonmantank (approved):

PR #366, opened 2020-10-02
Fixes for get records

tbedford:
# Description

See: 
* [DEVX-4547](https://nexmoinc.atlassian.net/browse/DEVX-4547).
* [DEVX-4548](https://nexmoinc.atlassian.net/browse/DEVX-4548).

# Fixes

* For get records, every response example showed the product as SMS. Fixed to show correct product for each tab.
* For get records, fixed direction for example responses so that they were correctly inbound/outbound as appropriate.
* For get records, fixed tabbing between Response panel and tabs in corresponding example panel.

# Checklist

### DIFF
--- a/definitions/reports.yml
+++ b/definitions/reports.yml
@@ -801,6 +801,7 @@ components:
         - MESSAGES
         - CONVERSATIONS
         - ASR
+      example: "SMS"  
     product_sms:
       type: string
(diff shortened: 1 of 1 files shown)


marklewin (approved):
Tested in the review app and all looks good 👍

PR #367, opened 2020-10-05
Reinstate Voice Name field, marked as deprecated

lornajane:
# Description

When the recommended TTS fields changed from `voice_name` to `language` and `style`, we completely removed the `voice_name` from the API description but it's still supported in the API itself. A more accurate way to describe the API would be to include the field but clearly mark it as deprecated.

(see also: an open issue to have our docs mark this more clearly. Other platforms should do it though https://github.com/Nexmo/nexmo-oas-renderer/issues/87)

# Fixes

* Keep the old voice name field docs for reference

# Checklist

### DIFF
--- a/definitions/voice.yml
+++ b/definitions/voice.yml
@@ -323,6 +323,8 @@ components:
           $ref: "#/components/schemas/language"
         style:
           $ref: "#/components/schemas/style"
+        voice_name:
+          $ref: "#/components/schemas/voice_name"
         loop:
           type: integer
           description: The number of times to repeat the text the file, 0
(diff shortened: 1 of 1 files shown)


bencgreenberg (commented):
I think we should continue this conversation at the next product specialists meeting. The beginning of it during the SDK call was quite fruitful, and its completion will bear directly on this PR.

If we do bring back visibility to the deprecated and no longer advised `voiceName` parameter in the OAS doc, we should find a way in the renderer to hide it, as it's truly not the preferred format going forward for new or existing customers.

lornajane:
Just looping back to check if there's more we should do before we merge this?

bencgreenberg (approved):
Looks good... the update to the OAS renderer has been merged, and a new version of Station with it will be cut and released early next week, so this will render on ADP soon.

PR #369, opened 2020-10-07
Describing errors in callbacks

lornajane:
We talked about how to represent various errors in a per-endpoint or per-callback fashion. I started trying to write down how we could use [OpenAPI Examples](http://spec.openapis.org/oas/v3.0.3.html#example-object) as a way to tackle this but an example is worth a thousand words so I ended up with a pull request showing just a couple of examples to give you an idea.

I also added enums for the error codes, hoping the codes would be findable in the docs or something ... not sure this helps.

Since our docs doesn't support examples like this yet, I made an example to show one approach (it also uses dropdowns, which made me think of the dropdowns vs tabs conversation this week too). Redoc can render static HTML and I can't attach HTML to PRs so I've committed that as well ... let's try not to merge it into the main project though!

I like whole examples for a few reasons. It means we can show more real examples, with just the fields that make sense and all showing related values. So we could show a facebook messenger incoming message with IDs, or a whatsapp one with numbers. Also there are other tools that expect and support this standard examples approach - redoc can be seen here but also Stoplight/prism can understand examples and return them on command, making testing of edge cases easier. I like to encourage people to mock the 429 response rather than actually overloading our servers ;)


mheap:
This matches my expectations and is how we've handled it in the past, but using `oneOf` rather than examples.

The biggest concern about showing per-endpoint/callback errors with a dropdown (or other switching mechanism) is that customers can't <kbd>cmd+f</kbd> and type in a code and have it show up. With that in mind, I think it might be worth showing both per-endpoint/callback errors to show specifically what each endpoint can return, plus an exhaustive list on the page for them to search and find (potentially with a column linking to which endpoints/callbacks return that error)

lornajane:
The findability is tricky because we'd have to expand so much to make it discoverable. Is there value in using the same approach to errors as we do for the ones in HTTP responses? I'm a little unclear (maybe I picked the wrong specs to look for examples) what we're providing here beyond what is already returned as the `reason` in the same payload as the status code. But if we returned a URL (like the problem formats have) that could give a lot more context, common explanations, etc - could that help with getting specific information?

mheap:
Returning a URL would definitely solve it. The ask for having a searchable index on the API ref page comes from older products that don't have that URI in the response or the callback (which also usually don't have a `reason` field)

lornajane:
@mheap Do you have an API in mind? Should we look at specific APIs to see if the idea still works? For this one I think we could add some extra parsing and break the examples out into a structure to display in the docs - but each API is different.

lornajane:
Closing due to inactivity

PR #375, opened 2020-10-30
Verify API: error response for new accounts

lornajane:
# Description

For new accounts (that haven't topped up yet), users can only use their own number for Verify testing. If they try to access another number, they need a message telling them what went wrong. Uses Code 29 to align with SMS API error codes.

# Fixes

* Error message when new users try to access non-permitted destinations with Verify API 

# Checklist

### DIFF
--- a/definitions/verify.yml
+++ b/definitions/verify.yml
@@ -3,7 +3,7 @@ servers:
   - url: "https://api.nexmo.com/verify"
 info:
   title: Verify API
-  version: 1.1.5
+  version: 1.1.6
   description: >-
     The Verify API helps you to implement 2FA (two-factor authentication) in your applications.
     This is useful for:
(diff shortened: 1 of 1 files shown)


lornajane:
Can you review but not merge please? The change isn't actually live yet

marklewin (approved):
LGTM @lornajane ... approved but not merged as requested

lornajane:
Needs rebasing and the version number updating onto whatever version is current when the feature goes live and this PR is ready to merge

grquintero:
Can we remove the reference to "whitelist" and instead use "permitted destinations" or rather, "non-permitted destinations"? I'd prefer to move that direction than use whitelist vs blacklist. Thank you.

tbedford:
Yes, writing style guide is to avoid using whitelist/blacklist etc.

PR #392, opened 2020-11-11
Rebranding for SMS API specs

niati:
# Description

Changed Nexmo references to Vonage.

# Checklist


marklewin (approved):
Looks good. I replaced some absolute URL doc links so that we don't have to worry about those when the domain changes to vonage.com

tbedford (approved):

PR #400, opened 2020-11-18
DEVX-2854 Update Voice Webhooks

vshisterov:
# New 

Added `connection_timeout` and `socket_timeout` to `voice` capabilities - webhooks.

# Checklist

### DIFF
--- a/definitions/application.v2.yml
+++ b/definitions/application.v2.yml
@@ -1,7 +1,7 @@
 ---
 openapi: "3.0.0"
 info:
-  version: 2.0.6
+  version: 2.0.7
   title: "Application API"
   description: |
     Vonage provides an Application API to allow management of your Vonage Applications.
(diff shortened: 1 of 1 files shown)


lornajane:
If I've understood this right, these timeout values are how long Vonage should wait for a response to a webhook or a websocket connection. I think the wording of the description could make this clearer to a user. Do these fields have default values, or what happens if the fields aren't supplied?

vshisterov:
> If I've understood this right, these timeout values are how long Vonage should wait for a response to a webhook or a websocket connection. I think the wording of the description could make this clearer to a user. Do these fields have default values, or what happens if the fields aren't supplied?

There is some explanation coming in this PR:

bencgreenberg (approved):

lornajane:
I see that this was merged but I still think that a more complete description field would be appropriate and not all of the fields have default values, for example in the section for modifying a call. Could we possibly revisit please?

PR #403, opened 2020-11-25
Tidy/Restructure Account API description

lornajane:
# Description

Restructure the Account API to have named request bodies along the same lines as our other APIs (and to please the tool I'm using)

# Fixes

* Improves structure of Account API description

# Checklist

### DIFF
--- a/definitions/account.yml
+++ b/definitions/account.yml
@@ -110,7 +110,7 @@ paths:
       operationId: changeAccountSettings
       summary: Change Account Settings
       description: >-
-        Update the default callback URLs (where the webhooks are sent to) associated with your account:
+        Update the default webhook URLs associated with your account:
           * Callback URL for incoming SMS messages
           * Callback URL for delivery receipts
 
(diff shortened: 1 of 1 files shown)


definitions/account.yml
```diff
@@ -630,6 +590,56 @@ components:
           description: Internal Trace ID
           example: bf0ca0bf927b3b52e3cb03217e1a1ddf
 
+    accountSettingsRequest:
+      type: object
+      properties:
+        moCallBackUrl:
+          description: >-
+            The URL where Vonage will send a webhook when an SMS is received
```
marklewin, line 598: I always wonder where this "sending a webhook" terminology comes from. Vonage doesn't "send" a webhook, it makes a request to the webhook?

marklewin (commented):
Looks good @lornajane - just one comment on the language (which I know isn't strictly the reason for the PR, but it would make sense to fix it here if we are in agreement?)

definitions/account.yml
lornajane: This isn't bothering me at all, you can definitely think of it as "sending a payload to a webhook endpoint". If we have standard wording I am of course happy to conform (or to have you fix that directly on this PR if you'd like)

marklewin: @lornajane Changed some of the wording around webhooks. Also, my understanding is that you cannot set a number-specific DLR webhook, so attempted to make that clear in the docs (`$accountSettingsRequest`)

lornajane: Oh this is good, thanks for taking the care to make this better!

lornajane:
I like your changes @marklewin but I can't approve my own PR so consider your additions approved, and I've re-requested your review

marklewin (approved):
Approving Lorna's approval of my changes :)

PR #408, opened 2021-01-18
specify request id required condition

bencgreenberg:
# Description

* Specifies that either `request_id` or `request_ids` is required when searching for verify request.

# Checklist


marklewin (approved):
LGTM @bencgreenberg

PR #409, opened 2021-01-25
changing tool for Facebook linking to the dasbhoard took

slorello89:
# Description

In December we added a new tool to manage the linking of Facebook accounts to your Vonage API account, this update reflects that. Also messenger.nexmo.com is being deprecated.

# Checklist


marklewin (approved):
Fixed typo and markdown list rendering, otherwise good to go

PR #415, opened 2021-02-02
Adding 403 and 409 to create WA deployment

slorello89:
# Fixes

* Add HTTP 403 and 409 error code documentation to the `/deployments` endpoint - addresses cases where user is trying to create a deployment for a number that already exists.


marklewin (approved):
LGTM

PR #422, opened 2021-04-07
Add concatenated field to sample response

marklewin:
# Description

Show "concatenated" field in response example

# Checklist


adamczopp (approved):
thanks, looks good

PR #425, opened 2021-04-14
VAPI Random From Number

vshisterov:
# Description
New VAPI feature added - the ability to use random `from` number, selected from one of the phone numbers assigned to the application.

# New 

* Add new `random_from_number` parameter to POST /v1/calls request;
* `from` in POST /v1/calls request is not mandatory now.

# Checklist

### DIFF
--- a/definitions/voice.yml
+++ b/definitions/voice.yml
@@ -1,7 +1,7 @@
 ---
 openapi: 3.0.0
 info:
-  version: 1.3.6
+  version: 1.3.7
   title: Voice API
   description:
     The Voice API lets you create outbound calls, control in-progress calls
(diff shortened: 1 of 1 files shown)


definitions/voice.yml
```diff
@@ -518,7 +518,6 @@ components:
       type: object
       required:
         - to
-        - from
```
abdulajet, line 521: I didnt know that from was becoming optional ?

vshisterov: it can't be mandatory now, because it can't be used together with `random_from_number: true`

abdulajet: sorry, I was getting confused between properties. worth copying the notice from the NCCO guide too about the `from` object

abdulajet (approved):
🚀

PR #427, opened 2021-04-16
Fix VAPI from object required status

abdulajet:
# Description

# Fixes

* Fixed the required status of the from object on the voice api specs

# Checklist


abdulajet:
Fabian looking into root cause

PR #433, opened 2021-05-11
CS Voice REST API

vshisterov:
# Description

Legs and Conversations endpoints added

# Checklist

### DIFF
--- a/definitions/voice.yml
+++ b/definitions/voice.yml
@@ -1,7 +1,7 @@
 ---
 openapi: 3.0.0
 info:
-  version: 1.3.6
+  version: 1.3.8
   title: Voice API
   description:
     The Voice API lets you create outbound calls, control in-progress calls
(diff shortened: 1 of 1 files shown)


definitions/voice.yml
```diff
@@ -79,42 +102,25 @@ paths:
...
-        "201":
-          description: Created
-          content:
-            application/json:
-              schema:
-                "$ref": "#/components/schemas/CreateCallResponse"
-  "/{uuid}":
+                "$ref": "#/components/schemas/GetLegsResponse"
+    
+  "/v0.3/legs/{uuid}":
```
dragonmantank, line 123: Is there a reason this route isn't product namespaced like a lot of other routes? Will `legs` have it's own API versioning outside of any other product?

definitions/voice.yml
```diff
@@ -201,21 +207,17 @@ paths:
...
-          description: Ok
-          content:
-            application/json:
-              schema:
-                "$ref": "#/components/schemas/StopStreamResponse"
+        "204":
+          description: No content
 
-  "/{uuid}/talk":
+  "/v0.3/legs/{uuid}/talk":
```
dragonmantank, line 220: Is the old route going away, or staying for backwards compatibility? If it's the latter, we need to deprecate the old routes instead of completely removing them as they still exist and customers are still using them. We have a policy to not remove a route until it's fully sunset.

definitions/voice.yml
```diff
@@ -294,7 +292,277 @@ paths:
...
+      parameters:
+        - name: beep
+          in: query
+          schema:
+            type: boolean
+      responses:
+        "204":
+          description: No content
+
+  "/v0.3/conversations/{conversation_uuid}/events":
```
dragonmantank, line 342: I see there are a number of conversation routes now being moved here versus the original Conversation API. Are these routes still considered Beta? Are there intentions to move them up to the same version number as VAPI itself?

We run into confusion already with `beta1`, `beta2`, `0.2`, etc, on Conversations and I'd rather not have that same confusion with VAPI as well.

dragonmantank (changes requested):
Questions around versioning and whether these routes are appropriate under the Voice spec or if they should still live in their own spec.

definitions/voice.yml
pardel: old route is being deprecated. Moving from VAPI v1 to CAPI v0.3

definitions/voice.yml
pardel: Voice part of CAPI is now effectively GA.

PR #434, opened 2021-05-13
APICORE-5132 new tiering changes to subaccount API spec

EvaHergar18:
# Description

# Checklist


definitions/subaccounts.yml
```diff
@@ -73,7 +73,16 @@ paths:
         "401":
           $ref: "#/components/responses/UnauthorizedError"
         "403":
-          $ref: "#/components/responses/ForbiddenError"
+          description: Forbidden Error
```
ryanatnexmo, line 76: I was hoping there would be a cleaner way to use multiple `$ref` to different responses... But I can see now that this is the way (thanks to [stackoverflow](https://stackoverflow.com/questions/47447403/how-to-define-different-responses-for-same-http-status-code-in-openapi-swagger/47453822))

definitions/subaccounts.yml
```diff
@@ -825,6 +861,50 @@ components:
...
+          type: string
+        instance:
+          type: string
+      required:
+        - type
+        - title
+        - detail
+        - instance
+  examples:
+    account-not-provisioned:
```
ryanatnexmo, line 881: I think that this is a duplicate of the existing `ForbiddenError`. Is it possible to have `ForbiddenError` use this schema? Then there's only one place to update and no need to worry about them drifting :)

ryanatnexmo (changes requested):

definitions/subaccounts.yml
EvaHergar18: Actually I think all error messages can follow this structure and we can add one schema and different examples for each

definitions/subaccounts.yml
```diff
@@ -825,6 +832,63 @@ components:
...
+          "title": "Maximum number of sub-accounts reached",
+          "detail": "This account has reached maximum number of allowed sub-accounts",
+          "instance": "158b8f199c45014ab7b08bfe9cc1c12c"
+        }
+    independent-balance-disabled:
+      summary: "Primary account does not allow independent balance"
+      value:
+        {
+          "type": "https://developer.nexmo.com/api-errors/account/subaccounts#validation",
+          "title": "Invalid operation for account",
```
ryanatnexmo, line 879: The title should be `Invalid use_primary_account_balance` for this error

PR #444, opened 2021-09-08
OTT-3927: MMS Vcard V0 specs

usama1020:
# Description

# Checklist


definitions/messages-olympus.yml
```diff
@@ -346,6 +355,70 @@ components:
...
+                type:
+                  type: string
+                  description: The type of message to be sent
+                  example: image
+                  enum:
+                    - image
+                image:
+                  type: object
+                  required:
+                    - url
```
AFavreNexmo, line 383: It could be caption as optional as well, isn't it?

definitions/messages-olympus.yml
```diff
@@ -346,6 +355,70 @@ components:
...
+                  example: vcard
+                  enum:
+                    - vcard
+                vcard:
+                  type: object
+                  required:
+                    - url
+                  properties:
+                    url:
+                      description: The URL of the vcard attachment. The vcard file is available for 48 hours after it is created.
```
AFavreNexmo, line 418: Where does this 48 hours come from?

AFavreNexmo (approved):
Looks good, would be good just to generate the Heroku to see how it looks, but I forgot how to do it.

definitions/messages-olympus.yml
usama1020: yh, it is missing. I'll add it.

definitions/messages-olympus.yml
usama1020: i'm not sure. We have it for mms-image message so i guessed it'll be same for vcard.

AFavreNexmo (approved):
Looks good to me

AFavreNexmo (approved):
Looks good to me 🚀

pardel (approved):

PR #445, opened 2021-09-10
rate-limit added

adamczopp:
consulted with SRE the value

# Description

# Checklist


niati (approved):

niati (commented):
Just need to update the version to 2.1.18

pardel:
@niati I would suggest creating an entry here: https://developer.nexmo.com/api-errors and direct users there as part of the API response. See how redact does it: https://nexmo-api-spec-pr-224.herokuapp.com/redact

niati:
> @niati I would suggest creating an entry here: https://developer.nexmo.com/api-errors and direct users there as part of the API response. See how redact does it: https://developer.nexmo.com/api/redact

That's a good suggestion, @pardel. I will add those references as soon as Adam's updates are merged. @adamczopp - just need to bump up the API spec version number in the reports.yml file to get this PR approved and merged.

PR #447, opened 2021-09-22
Update Meetings API specification

ahinoamt:
# Description

# Checklist


ahinoamt:
For some reason, I can't add you as a reviewer, so I'm asking for your review here :) @abadyan-vonage

definitions/meetings.yml
```diff
@@ -96,6 +96,17 @@ paths:
                         room_id:
                           type: string
                           format: uuid
+                        room_type:
+                          type: string
+                          enum:
+                            - instant
+                            - long_term
+                        expiration_date:
```
abadyan-vonage, line 104: Our convention so far has been that the date fields are called "XXX_at" and their format is an ISO-8601 string.
So I think this should be called "expiring_at" and "created_at".
Regarding the format, we discussed switching to epoch time - If this is the final product decision, let's convert all exisiting datetime fields to this format to avoid confusion.
Also, shouldn't the type be `number` if we're using epoch format?

abadyan-vonage (changes requested):

definitions/meetings.yml
pardel: @abadyan-vonage all our APIs are using ISO8601 for dates so please use that format. 
https://nexmo.github.io/api-standards/standard/dates-and-times/

ahinoamt: About the naming convention- we had several meetings about it, we've decided about the naming 'XXX_date' (you can see it here under 'Create Room Response' -> https://docs.google.com/document/d/1SdA65ehCFjAPCdg43qQspAiMVO1UG81Mk_V1H8V9ywA/edit#heading=h.vqirb474yg8z)

About the time format- we agreed to stay consistent with Tokbox API, they represent the dates in UNIX epoch format (timestamp in milliseconds), you can see it here -> https://confluence.vonage.com/display/MEET/API+Guidelines

About the timestamp type (`string` instead of `number`)- the only reason that I set it like this is that Tokbox also represents it as `string` in their API and I didn't find another approach. I can change it to `number` if you believe it's better. 
screenshot from opentok API ->
Screen Shot 2021-09-23 at 14 22 42
 
Do we have more dates that need to be changed according to this convention?

pardel: @ahinoamt @abadyan-vonage The APIs across all products need to be consistent with the current Core APIs standards. As the TokBox APIs will be slowly integrated in the Core, things will change so I would not use the TokBox as the standard. 
FYI, DevRel needs to be involved into the design of the APIs as well.

ahinoamt: @pardel @abadyan-vonage I've changed the dates (name and format) according to Nexmo API standard as you requested

abadyan-vonage (approved):
LGTM

PR #454, opened 2021-11-08
Feature/number insight sync realtime

SecondeJK:
# Description

The Numbers Insight API now offers a new parameter for advanced accounts. An API request can now take the `real_time_data` boolean parameter, which will render an object named `real_time_data` with two keys: `active_status` and `handset_status`.

Changes visible at: https://nexmo-api-spec-pr-454.herokuapp.com/number-insight

# Checklist

### DIFF
--- a/definitions/number-insight.yml
+++ b/definitions/number-insight.yml
@@ -109,7 +109,7 @@ paths:
                 content:
                   application/json:
                     schema:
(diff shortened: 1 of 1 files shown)


pardel (changes requested):
The async part should be added under a Webhooks part. See https://developer.vonage.com/api/messages-olympus

pardel (approved):

PR #455, opened 2021-11-09
reverted messages v1

pardel:
reverted https://github.com/Nexmo/api-specification/pull/453 as it's a duplicate


superchilled (approved):
Looks good

PR #469, opened 2022-01-19
added rate limit and bumped the version

adamczopp:
moving the comments over:
@niati I would suggest creating an entry here: https://developer.nexmo.com/api-errors and direct users there as part of the API response. See how redact does it: https://developer.nexmo.com/api/redact

# Description

# Checklist

### DIFF
--- a/definitions/reports.yml
+++ b/definitions/reports.yml
@@ -1,6 +1,6 @@
 openapi: '3.0.0'
 info:
-  version: 2.2.0
+  version: 2.2.1
   title: Reports API
   description: |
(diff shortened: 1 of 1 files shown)


definitions/reports.yml
```diff
@@ -2357,3 +2357,6 @@ x-errors:
     link:
       text: View API reference
       url: /api/reports#cancel-report
+   rate-limit:
```
niati, line 2360: @adamczopp it looks like on line 2360 that you may have one too many spaces before 'rate-limit' . We think that's what's causing the error.

niati (approved):

PR #477, opened 2022-03-03
Fix response status for GET v1/calls

amdcavallaro:
# Description

# Fixes

Fixes the status showing from started to complete

https://developer.vonage.com/api/voice?theme=dark#getCalls


abdulajet (approved):
Reviewed with victor on a call

PR #502, opened 2022-06-20
change to number insight, return value for async realtime was wrong type

SecondeJK:
# Description

The original Number Insight changes to add `real_time_data` were incorrect for the return value. `active_status` was listed as a `bool`, but is actually a `string` with three possible values.

Example:

# Fixes

* Fixed return value and type description for `real_time_data`
-->

# Checklist


dragonmantank (approved):
LGTM

SMadani (approved):

PR #511, opened 2022-07-28
Added callout for network unblock

SecondeJK:
# Description

By request of product, we need a callout to let developers know that the network unblock feature is only available
on request.

# New 

* Added callout to Network Unblock section of Verify v1 under Fraud Management.

# Checklist


definitions/verify.yml
```diff
@@ -521,7 +521,16 @@ paths:
     post:
       operationId: networkUnblock
       summary: Request a network unblock
-      description: Request to unblock a network that has been blocked due to potential fraud detection
+      description: >-
+        Request to unblock a network that has been blocked due to potential fraud detection
+
+        <div class="Vlt-callout Vlt-callout--critical">
+        <i></i>
```
SMadani, line 528: Why is this empty?

pardel: Can we also avoid using HTML tags in here?

superchilled (changes requested):
Check failing for version bump.

Also see my comment re line 528

superchilled (approved):

PR #513, opened 2022-07-28
DEVX-6212/DEVX-6276/DEVX-6392 Messages spec updates for WhatsApp

superchilled:
# Description

## New 

* Adds `whatsapp` object to Message Status callbacks for WhatsApp. This is as a result of WhatsApp's change to conversation-based pricing rather than per-message pricing. The  `whatsapp` object contains data about the conversation to which the message relates.

## Updated

* Updates the description of the `price` property in the `usage` object for Message Status callbacks for WhatsApp. Again, this is to reflect the move to conversation-based pricing for WhatsApp messages.

# Checklist


definitions/messages-olympus.yml
```diff
@@ -1,6 +1,6 @@
 openapi: "3.0.1"
 info:
-  version: 1.1.1
+  version: 1.2.0
```
SecondeJK, line 3: With the change here, are we definitely not creating any backwards compatibility issues if the structure of the pricing payload is different?

definitions/messages-olympus.yml
```diff
@@ -1430,6 +1430,39 @@ components:
...
                 - read
+            usage:
+              type: object
+              properties:
+                currency:
+                  type: string
+                  example: EUR
+                  description: The charge currency in ISO 4217 format.
+                  enum:
+                    - EUR
```
SecondeJK, line 1441: this enum covers all values it can be, right? So you can't be charged in the messages API un USD?

definitions/messages-olympus.yml
superchilled: It's not changing anything existing. It's adding a node in the callback that wasn't there before.

definitions/messages-olympus.yml
superchilled: Messages are pre-paid via the Dashboard, so I'm pretty sure it's all charged in EUR.

SecondeJK (approved):

PR #520, opened 2022-08-09
Fixed create session's content type

dragonmantank:
Changed to `application/x-www-form-urlencode`

No version bump as this was a typo


SMadani (approved):

maxkahan (commented):
application/x-www-form-urlencoded or application/x-www-form-urlencode?

PR #522, opened 2022-08-16
setup tools to run validation and rendering locally

dorr-vonage:
This is not an API change.

Setup local tools to run validation and rendering locally.
I've updated the readme with suggested changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,27 +35,58 @@ These definitions provide a single point of truth that can be used end-to-end:
 
 We use [Spectral](https://stoplight.io/open-source/spectral/) to style and validity check our files. If you're using Stoplight Studio, this picks up our style definition by default. Otherwise, you can install the tool locally like this:
 
-`npm install -g @stoplight/spectral`
+Install node version. If using node version manger you can:
+```
+nvm install && nvm use
+```
 
-And check your file (for example `verify.yml`):
+Install node dependencies:
+```
+npm i
+```
 
-`spectral lint definitions/verify.yml`
+Run validation:
+```
+npm run validate
+```
 
 This runs the same checks that we use in the build that runs when you open a pull request.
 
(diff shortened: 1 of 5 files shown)


dorr-vonage:
@Nexmo/devrel I would appreciate it if anyone could review this PR and provide any comments or approval.

manchuck (approved):
It looks good, but do we need to instruct users on installing NodeJS?

dorr-vonage:
@manchuck thanks for your review. I did change the instruction to suggest installing NodeJs if not already available.
Should I merge to main or this is done by the devrel team?

dorr-vonage:
OK, I just notice I don't have permission to merge "The base branch restricts merging to authorized users"

README.md
```diff
@@ -35,9 +35,9 @@ These definitions provide a single point of truth that can be used end-to-end:
 
 We use [Spectral](https://stoplight.io/open-source/spectral/) to style and validity check our files. If you're using Stoplight Studio, this picks up our style definition by default. Otherwise, you can install the tool locally like this:
 
-Install node version. If using node version manger you can:
+Install NodeJS, if not already installed. If using node version manger (nvm) you can:
 ```
-nvm install && nvm use
+nvm use || nvm install
```
manchuck, line 40: You should return to the previous operator. If `nvm use` fails, then `nvm install` will also fail

manchuck (changes requested):

README.md
dorr-vonage: if the version specified in .nvmrc is already installed then use will work. Otherwise, nvm will install and set the .nvmrc version

pardel: @dorr-vonage we have actually migrated from this repo and we use a different renderer. Will DM you the info.

PR #525, opened 2022-08-28
Modified application v2

ruskibenya:
# Description

Added Video API capabilities to the Application Layer. Users can now create and view applications with the video capabilities with the `GET /` and `POST /` endpoints. Users can retrieve applications and edit applications to enable video with the 'GET /{id}'. And they can delete with `DELETE /{id}`. The `ApplicationResponse`model, at reference #/components/schemas/ApplicationResponse",  has been updated accordingly.

@dragonmantank can you look over that this was done right? Also, would this be a major or minor change?

### DIFF
--- a/definitions/application.v2.yml
+++ b/definitions/application.v2.yml
@@ -1,22 +1,21 @@
----
-openapi: "3.0.0"
+openapi: 3.0.0
 info:
   version: 2.1.1
-  title: "Application API"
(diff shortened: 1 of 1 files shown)


dragonmantank:
Looks OK so far, but not approving until we know if any additional add-ons are being put into Application config since the focus change.

PR #528, opened 2022-09-02
OTT-4452: Remove WA template namespace

alberto-brigandi:
it's not mandatory anymore.
Also changed the example values to use a pre-populated WA template

# Description

# Fixes

* Removed WhatsApp template namespace since it's not required anymore
* Improved the WhatsApp template example to use a template pre-populated in all embedded sign-up accounts

# Checklist

### DIFF
--- a/definitions/messages-olympus.yml
+++ b/definitions/messages-olympus.yml
@@ -567,8 +567,6 @@ components:
       type: array
       description: The parameters are an array of strings, with the first string being used for {{1}} in the template, with the second being {{2}} etc. Only required if the template specified by `name` contains parameters.
       example:
-        - Verification
-        - '2526'
         - '5'
       items:
         type: string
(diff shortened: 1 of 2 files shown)


peterjurkovic (approved):

definitions/messages-olympus.yml
```diff
@@ -567,8 +567,6 @@ components:
       type: array
       description: The parameters are an array of strings, with the first string being used for {{1}} in the template, with the second being {{2}} etc. Only required if the template specified by `name` contains parameters.
       example:
-        - Verification
```
AFavreNexmo, line 570: The description is showing param {2}, it might worth leaving 2 in the array. Maybe more explicit like myFirstParameter, mySecondParameter

AFavreNexmo (approved):

definitions/messages-olympus.yml
alberto-brigandi: It'd be nice, but there's no pre-populated template with 2 parameters that can be used with message type template. All the other ones need for a media to be specified as header, which we don't support and it requires to use custom type

LeonPatmore (approved):

pardel:
@alberto-brigandi can we find a way to either merge this in or get rid of it please?

PR #530, opened 2022-09-06
added apphash and client ref to webhooks

SecondeJK:
# Description

This PR adds `client-ref` into the webhook payloads for Verify V2 and adds the `app_hash` parameter when requesting a Verify as per request here: https://jira.vonage.com/browse/APIDOC-391

Example:

# New 

* Added `client-ref` to Verify V2 Webhook response payloads as optional.
* Added `app_hash` as optional request parameter when sending a new SMS Verify request

# Checklist


superchilled (approved):
LGTM

PR #535, opened 2022-09-16
Update Context Field Desc for the Messages API (Whats App)

mbcrump:
# Updated

* Added a description for the Context object to make it more clear for end users to know what it does. The description added was: This is only present for the Inbound Message where the user is quoting another message. It provides information about the quoted message.


definitions/messages-olympus.yml
```diff
@@ -1,6 +1,6 @@
 openapi: "3.0.1"
 info:
-  version: 1.2.2
+  version: 1.2.1
```
superchilled, line 3:
```suggestion
  version: 1.2.3
```

Not sure why we're decrementing the patch version here. We probably want to increment it.

superchilled (approved):

PR #560, opened 2022-11-13
Add total_items to get rooms request

elishaikatz:
# Description

# Checklist

### DIFF
--- a/definitions/meetings.yml
+++ b/definitions/meetings.yml
@@ -24,6 +24,10 @@ paths:
                     type: integer
                     example: 10
                     description: The number of results returned on this page.
+                  total_items:
+                    type: integer
+                    example: 30
+                    description: The overall number of available rooms.
                   _embedded:
                     type: object
                     properties:
--- a/definitions/meetings.yml
+++ b/definitions/meetings.yml
@@ -1,7 +1,7 @@
 openapi: 3.0.0
 info:
   title: Vonage Meetings Public API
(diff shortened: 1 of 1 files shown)


pardel:
@elishaikatz would it be possible to have a cursor-based pagination rather than page-based? See https://nexmo.github.io/api-standards/standard/pagination/ for our standars.

elishaikatz:
> @elishaikatz would it be possible to have a cursor-based pagination rather than page-based? See https://nexmo.github.io/api-standards/standard/pagination/ for our standars.

It's doable, but it's a big breaking change now, why do you ask? we created the current pagination based on the page-based from the Nexmo standards

pardel:
@elishaikatz our standards read: "APIs shall use a cursor as their pagination option unless otherwise agreed with DevRel."
https://nexmo.github.io/api-standards/standard/pagination/

elishaikatz:
> @elishaikatz our standards read: "APIs shall use a cursor as their pagination option unless otherwise agreed with DevRel." https://nexmo.github.io/api-standards/standard/pagination/

@pardel  As I know it, this feature was implemented a long time ago, and we are only adding now the total items property - unfortunately, Alex from my team, that implemented this feature, is no longer working at Vonage, so I cannot ask who he worked with a while ago with dev rel, but this was approved - as Meetings API basic proposal was approved.

pardel (approved):

PR #565, opened 2022-11-18
DEVX-6530: SMS Spec update to removing unsupported types

superchilled:
# Description

There are some message types in the SMS API that are no longer supported. Updating the spec to remove these types and any associated properties.

Example:

# Updated

* `type` enum: removed the `wappush`, `vcard`, and `vcal` options

# Removed

* `vcard`
* `vcal`
* `title`
* `url`
* `validity`

# Checklist


SMadani (approved):

SMadani:
Presumably a notice is out in the main docs about this?

superchilled:
> Presumably a notice is out in the main docs about this?

I checked the docs, and we don't even mention these message types. From discussion with the PM, very few customers actually use them, and those that do have apparently been directly advised.

PR #567, opened 2022-11-21
Added /enum endpoint (Apidoc 519)

EDDYMENS:
# Description

# New 

* Added `/enum` endpoint to provide an up-to-date list of all available options for enum parameters 

# Checklist


definitions/10dlc.yml
```diff
@@ -545,6 +545,26 @@ paths:
       responses:
         "200":
           $ref: "#/components/responses/VettingProviders"
+  /enum:
```
pardel, line 548: Can we rename this to '/values'?

definitions/10dlc.yml
```diff
@@ -545,6 +545,26 @@ paths:
...
+      operationId: getEnumValues
+      responses:
+        '200':
+          description: List of all enumeration values
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  type: string
```
pardel, line 563: Can we add an example here?

definitions/10dlc.yml
```diff
@@ -2957,6 +2977,8 @@ tags:
     description: APIs relating to working with Campaigns
   - name: Numbers
     description: APIs relating to working with Numbers in Campaigns
+  - name: Enums
+    description: The list of options available for enum properties across all APIs
```
pardel, line 2981: "options available" => "values possible"

pardel (changes requested):

definitions/10dlc.yml
pardel: 👍

definitions/10dlc.yml
EDDYMENS: Example added

definitions/10dlc.yml
EDDYMENS: Rewording applied

definitions/10dlc.yml
```diff
@@ -1416,6 +1430,22 @@ components:
...
+      description: List of all enumeration values
+      content:
+        application/json:
+          schema:
+            type: object
+            properties:
+              entity_type:
+                type: array
+                items:
+                 type: object
```
pardel, line 1443: Are you sure this is right? I would expect an array of values not of objects.

definitions/10dlc.yml
pardel: I'm not seeing the change...

pardel (changes requested):

definitions/10dlc.yml
EDDYMENS: @pardel yep it's correct, as per the output I got from my test.
Screenshot 2022-11-23 at 15 11 06

definitions/10dlc.yml
EDDYMENS: Had it in two places. It should show up now.

pardel (approved):

PR #586, opened 2023-02-03
Updated Video API spec with Broadcast and SIP endpoints

dragonmantank:
# Description

## New
Added the Broadcast and SIP endpoints to the Video API spec for the Video Integration

## Changed
A bunch of quotes because this is one of the first specs to get run through Stoplight Studio

# Checklist

### DIFF
--- a/definitions/video.yml
+++ b/definitions/video.yml
@@ -1,16 +1,16 @@
 openapi: 3.0.3
 x-stoplight:
-  id: dka9bkyfsi8b4
+  id: 3z2phq4skuv6u
 info:
   title: Vonage Video
-  version: "0.2.1"
+  version: 0.3.0
   contact:
     name: Vonage DevRel
-    url: "https://developer.vonage.com"
+    url: 'https://developer.vonage.com'
     email: devlre@vonage.com
   description: APIs relating to Vonage Video
 servers:
-  - url: "https://video.api.vonage.com"
+  - url: 'https://video.api.vonage.com'
     description: Vonage Video
 paths:
   /session/create:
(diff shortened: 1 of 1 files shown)


SMadani:
Would it make sense to introduce dividers for SIP and Broadcast for consistency? Currently there is no section for them, they're just at the top but not on the navigation menu / jump links.

dragonmantank:
@SMadani Added tags for those sections

definitions/video.yml
```diff
@@ -889,37 +876,592 @@ paths:
...
+              schema:
+                type: object
+                properties:
+                  id:
+                    type: string
+                    description: A unique ID for the SIP call.
+                    example: b0a5a8c7-dc38-459f-a48d-a7f2008da853
+                  connectionId:
+                    type: string
+                    description: 'The OpenTok connection ID for the SIP call''s connection in the OpenTok session. You can use this connection ID to terminate the SIP call, using the OpenTok REST API.'
```
SMadani, line 931: Just wondering if we're keeping the "OpenTok" naming for descriptions for the time being?

SMadani:
Should "count" in "List Live Streaming Broadcasts" be a string? The example shows "abc123", I'm guessing it's meant to be an integer wrapped in a string.

SMadani:
`status` values for broadcasts are listed [here](https://tokbox.com/developer/rest/#get_info_broadcast). Must be one of `started` or `stopped`. RTMP status must be one of `connecting`, `live`, `offline` or `error`.

SMadani:
Some of the fields for the Broadcast request / response are missing documentation, which can be found [here](https://tokbox.com/developer/rest/#start_broadcast). The "All fields" tab is missing `maxBitrate` in the example request.

SMadani:
Streams, Application, Broadcast, Archive IDs are actually GUIDs. It seems to be only Sessions that are proprietary (i.e. not compliant with the UUID spec). In strongly typed SDKs they can thus be represented by a UUID class rather than a String. Same for the URL / URI fields.

SMadani:
Broadcast should not have a `streams` property in the response. Also, changing the broadcast layout should be PUT, not POST. `outputs.rtmp` is not required; the logic is that there must be HLS or at least one RTMP (or both), but RTMP is not mandatory if HLS is specified. On that note, the example for Required fields for creating a broadcast only shows session ID. It should have outputs as well.

definitions/video.yml
dragonmantank: Updated

dragonmantank:
Ultimately moving this discussion to https://github.com/Vonage/bifrost-dev-docs-mfe/pull/370, but:

> In the `/dial` endpoint, shouldn't the `sip` object be required? Since the URL is required. Or does it mean that it is the only required property if `sip` is specified? Also, the "All Fields" example is missing the three Boolean parameters.

Correct. The old spec didn't have it, but `sip.uri` is a hard requirement so making that explicitly required.

> Should "count" in "List Live Streaming Broadcasts" be a string? The example shows "abc123", I'm guessing it's meant to be an integer wrapped in a string. Also, should offset and count be strings in the request body?

That was a discrepancy between the docs and the spec. So when I pasted their example response in, they had it as a string. 

> `status` values for broadcasts are listed [here](https://tokbox.com/developer/rest/#get_info_broadcast). Must be one of `started` or `stopped`. RTMP status must be one of `connecting`, `live`, `offline` or `error`.

Updated

> 

Should be updated now

> Streams, Application, Broadcast, Archive IDs are actually GUIDs. It seems to be only Sessions that are proprietary (i.e. not compliant with the UUID spec). In strongly typed SDKs they can thus be represented by a UUID class rather than a String. Same for the URL / URI fields.

I'm inclined to leave up the deserialization for this to the SDK rather than enforce in the API spec. I can't guarantee those will be compliant with any specific UUID version (of which there are multiple). 

> Broadcast should not have a `streams` property in the response. Also, changing the broadcast layout should be PUT, not POST. `outputs.rtmp` is not required; the logic is that there must be HLS or at least one RTMP (or both), but RTMP is not mandatory if HLS is specified. On that note, the example for Required fields for creating a broadcast only shows session ID. It should have outputs as well.

I'm not sure why Broadcast should not have a `streams` property in the response. It's an array that is always returned, even if empty.

I removed the requirement for `outputs.rtmp` on new broadcasts. We don't have a way to really show "required but only if conditions are met". There is a janky workaround but our renderer doesn't seem to support it.

